PR #1, opened 2016-07-29
Add a github service with issue/PR expansion

Kegsay:
This adds a Github service which will expand issues that look like `owner/repo#11`. It uses an unauthenticated client for now (so no database required!), and is mainly so I can get to grips with the Go-NEB plugin architecture.

### DIFF
--- a/src/github.com/matrix-org/go-neb/services/github/github.go
+++ b/src/github.com/matrix-org/go-neb/services/github/github.go
@@ -14,7 +14,7 @@ import (
 
 // Matches alphanumeric then a /, then more alphanumeric then a #, then a number.
 // E.g. owner/repo#11 (issue/PR numbers) - Captured groups for owner/repo/number
-const ownerRepoIssueRegex = "([A-z0-9-_]+)/([A-z0-9-_]+)#([0-9]+)"
+var ownerRepoIssueRegex = regexp.MustCompile("([A-z0-9-_]+)/([A-z0-9-_]+)#([0-9]+)")
 
 type githubService struct {
 	id     string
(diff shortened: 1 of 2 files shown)


src/github.com/matrix-org/go-neb/services/github/github.go
```diff
@@ -0,0 +1,99 @@
...
+func (s *githubService) ServiceUserID() string { return s.UserID }
+func (s *githubService) ServiceID() string     { return s.id }
+func (s *githubService) ServiceType() string   { return "github" }
+func (s *githubService) RoomIDs() []string     { return s.Rooms }
+func (s *githubService) Plugin(roomID string) plugin.Plugin {
+	return plugin.Plugin{
+		Commands: []plugin.Command{},
+		Expansions: []plugin.Expansion{
+			plugin.Expansion{
+				Regexp: regexp.MustCompile(ownerRepoIssueRegex),
```
NegativeMjark: I'd prefer to have the regexp stored as a top-level variable, That way the MustCompile will fail earlier if the rexexp is bad. (Also means that we know that we will only construct the regexp once.)

src/github.com/matrix-org/go-neb/services/github/github.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	httpCli := oauth2.NewClient(oauth2.NoContext, tokenSource)
+	return github.NewClient(httpCli)
+}
+
+// ownerRepoNumberFromText parses a GH issue string that looks like 'owner/repo#11'
+// into its constituient parts. Returns: owner, repo, issue#.
+func ownerRepoNumberFromText(ownerRepoNumberText string) (string, string, int, error) {
+	// TODO: cache this?
+	re := regexp.MustCompile(ownerRepoIssueRegex)
```
NegativeMjark: Again, I'd prefer it if the regexp was stored at the top-level.

Kegsay: done

src/github.com/matrix-org/go-neb/services/github/github.go
Kegsay: done

NegativeMjark:
LGTM

PR #2, opened 2016-08-01
Add a webhook handler and parse incoming Github webhook events

Kegsay:
- Does not persist anything yet.
- Does not send messages to Matrix rooms yet.
- Add tests for Github webhook event parsing. (most of the diff)


NegativeMjark:
LGTM

PR #3, opened 2016-08-03
Add skeleton for auth modules

Kegsay:
`AuthModules` are interfaces which implement auth for a resource, much like how `Services` are interfaces which implement a service for a user.

This PR adds support for `AuthModules` along with storage in a new table `third_party_auth`. This is exposed via a `POST` API `/admin/configureAuth`.

### DIFF
--- a/src/github.com/matrix-org/go-neb/api.go
+++ b/src/github.com/matrix-org/go-neb/api.go
@@ -16,6 +16,32 @@ func (*heartbeatHandler) OnIncomingRequest(req *http.Request) (interface{}, *err
 	return &struct{}{}, nil
 }
 
+type configureAuthHandler struct {
+	db *database.ServiceDB
+}
+
+func (*configureAuthHandler) OnIncomingRequest(req *http.Request) (interface{}, *errors.HTTPError) {
+	if req.Method != "POST" {
+		return nil, &errors.HTTPError{nil, "Unsupported Method", 405}
+	}
+	var tpa types.ThirdPartyAuth
+	if err := json.NewDecoder(req.Body).Decode(&tpa); err != nil {
+		return nil, &errors.HTTPError{err, "Error parsing request JSON", 400}
+	}
+
+	am := types.GetAuthModule(tpa.Type)
+	if am == nil {
+		return nil, &errors.HTTPError{nil, "Bad auth type: " + tpa.Type, 400}
+	}
+
+	err := am.Process(tpa)
+	if err != nil {
+		return nil, &errors.HTTPError{err, "Failed to persist auth", 500}
+	}
+
+	return nil, nil
+}
+
 func handleWebhook(w http.ResponseWriter, req *http.Request) {
 	segments := strings.Split(req.URL.Path, "/")
 	// last path segment is the service type which we will pass the incoming request to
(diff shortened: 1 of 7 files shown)


NegativeMjark:
Overall this is looking pretty reasonable.

A couple of thoughts:

1) How do we handle jira? Do the auth modules themselves need to have IDs?

2) Should the database be a global variable? Everything wants to access it, there really should only be one of them, and it would save a lot of boilerplate everywhere.

Kegsay:
> 1) How do we handle jira? 

Via the `resource`. `AuthModules` are keyed off the `user_id` and the `resource` - the `resource` is there to provide the domain such that the JIRA auth module knows what to talk to e.g. `matrix.org/jira`.

> Do the auth modules themselves need to have IDs?

I thought hard about this, and we may or may not need them. The intended flow is:
 - User submits `/configureAuth` request.
 - `AuthModule` is loaded via the `type`.
 - Said module may persist a `third_party_auth` entry (e.g. to remember OAuth2 state vars) and return a URL / something for the user to do. This should clobber based on the tuple of `(user_id, resource)` so one user can only have 1 in-flight auth process in play at any one time.
 - At some point later in time, NEB is hit with a redirect request (how this URL is formed depends on the auth module).
 - NEB loads the `third_party_auth` entry *somehow* (for github you would do this based off the `state` param, for JIRA you want to do this off the tuple of `(resource,  state)` and finishes the auth.
 - NEB returns succes/failure.

We could make an ID field and it could be used as a stand-in for the `state` param which a lot (all?) OAuth endpoints require. The only problem with this is that we don't actually want to key off / clobber based on this, we want uniqueness to be based off `(user_id, resource)`. We can't do `UNIQUE(user_id, resource, id)` because that then allows multiple `(user_id, resource)` entries with different `ids`.

EDIT: Come to think of it, we probably actually want the tuple to be `(id, resource)` so we can ensure that no 2 users can get the same generated `state` param which we then cannot unpick on the redirect request.

> Should the database be a global variable? Everything wants to access it, there really should only be one of them, and it would save a lot of boilerplate everywhere.

I'd prefer it if things were explicit with what they need in order to do their job.

The idea that any file in the project can suddenly start doing database queries is a little terrifying. Not *everything* needs access to the database (echo service, heartbeat handler, webhook code), and being able to indicate which bits don't clearly to the reader is nice.

Making it global would also remove any usefulness that may come out by splitting up `ServiceDB` into smaller `structs`. For example, this auth code doesn't actually use *anything* from the `ServiceDB` struct, it's just there because it's easier since that object is available to other parts of the program already. If we had an `AuthModuleDB` in addition to the `ServiceDB`, then we could clamp down even further here and only pass an `AuthModuleDB` to an `AuthModule` which contains a subset of valid operations which `AuthModules` can perform on the database. Said `AuthModuleDB` will still be in the same package though (to get access to that delicious `runTransaction`).

NegativeMjark:
> The idea that any file in the project can suddenly start doing database queries is a little terrifying. Not everything needs access to the database (echo service, heartbeat handler, webhook code), and being able to indicate which bits don't clearly to the reader is nice.

Won't you learn what needs access to the database by seeing which things import the database module?

Kegsay:
> Won't you learn what needs access to the database by seeing which things import the database module?

Rarely are files that finely scoped though. Take `api.go`, which has all the HTTP handlers as an example.

NegativeMjark:
LGTM in so far as it will all be rewritten by subsequent work.

PR #4, opened 2016-08-03
Use the service ID in the webhook path instead of the service type

Kegsay:
This means we can load the `Service` from the database then invoke
`OnIncomingWebhook` on it, rather than load up a new Service.


NegativeMjark:
LGTM

PR #5, opened 2016-08-03
Send GH webhook notifications based on GH service config

Kegsay:
- The github service config now has a map of `room_id => [notification, event, types]`
- When a webhook is received, this map is checked against the event type in the header of the webhook request and is then sent via the service's matrix client.


NegativeMjark:
LGTM

PR #6, opened 2016-08-04
Add concept of AuthRealms

Kegsay:
- These represent a place where a user can authenticate themselves.
- They function in the same way as Services (insert/update based on an HTTP API)
- They currently don't *do* a lot other than exist for storing realm-specific
  information (e.g. the `GithubRealm` stores the `ClientSecret` and `ClientID`)

Most of the code is direct copypasta from `Services`. I've replaced the third-party auth endpoint with the realm endpoint for now. Another PR will remove `ThirdPartyAuth` entirely. Another PR after that will add in the concept of `AuthSession` which represents an individual session for an auth realm.


NegativeMjark:
LGTM

PR #8, opened 2016-08-04
Add concept of AuthSessions

Kegsay:
Auth sessions are a single auth process between a user and an auth realm. As
such, they are keyed off the tuple of `(user_id, realm_id)`.

Only the realm which they belong to knows how to construct them, hence all
"load" sections require an `AuthRealm` to be extracted first.

Currently I pass in a `json.RawMessage` rather than factory initialise and
clobber public fields based on the JSON, we can always change that if need be
later down the line.

Overall, this feels really nice (when starting to add in GH auth, everything I
wanted was already there in the right place waiting for me).

### DIFF
--- a/src/github.com/matrix-org/go-neb/database/schema.go
+++ b/src/github.com/matrix-org/go-neb/database/schema.go
@@ -43,6 +43,15 @@ CREATE TABLE IF NOT EXISTS auth_realms (
 	time_updated_ms BIGINT NOT NULL,
 	UNIQUE(realm_id)
 );
+
+CREATE TABLE IF NOT EXISTS auth_sessions (
+	realm_id TEXT NOT NULL,
+	user_id TEXT NOT NULL,
+	session_json TEXT NOT NULL,
+	time_added_ms BIGINT NOT NULL,
+	time_updated_ms BIGINT NOT NULL,
+	UNIQUE(realm_id, user_id)
+);
 `
 
 const selectServiceUserIDsSQL = `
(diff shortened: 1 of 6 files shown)


src/github.com/matrix-org/go-neb/database/schema.go
```diff
@@ -268,3 +277,64 @@ func updateRealmTxn(txn *sql.Tx, now time.Time, realm types.AuthRealm) error {
...
+	var realmType string
+	var sessionJSON []byte
+	row := txn.QueryRow(selectAuthSessionSQL, realmID, userID)
+	if err := row.Scan(&realmType, &sessionJSON); err != nil {
+		return nil, err
+	}
+	realm := types.CreateAuthRealm(realmID, realmType)
+	if realm == nil {
+		return nil, fmt.Errorf("Cannot create realm of type %s", realmType)
+	}
```
NegativeMjark: Might be worth plucking out the auth realm JSON from the database to fillout the common fields before calling AuthSession.

NegativeMjark:
What's the intended lifetime of a AuthSession? Does it contain the token? or is it just temporary bit of state for getting a token?

src/github.com/matrix-org/go-neb/database/schema.go

Kegsay:
The intention is that the `AuthSession` will contain the token. If a session exists for a user ID / realm ID tuple, then either they have successfully authed or they are in the process of authing (only the realm knows).

NegativeMjark:
Okay. Looks good enough to me to make an attempt at actually implementing the github auth. It'll be be clearer whether this API shape makes sense then.

PR #9, opened 2016-08-04
Tweak the auth session API to be more transactional

Kegsay:
- Rename the path from /configureAuthSession to /requestAuthSession
- Add a global getter/setter for the `ServiceDB` : this avoids cyclical deps
  because now the Realm wants access to the database, and due to the factory
  pattern it would mean `types.go` would need to import `database`, but
  `database` is already doing so to invoke the factory function in `schema.go`.
- Modify how `AuthSession` is loaded/stored in the database. Now it is just
  a blunt JSON store for Public fields. It is initialised via a new Realm
  interface function `AuthSession(userID, realmID)` which is there to return
  the right `struct` so stuff can be unmarshalled into it.
- Add a new Realm interface function `RequestAuthSession` which is invoked
  when `/requestAuthSession` is hit. It is a direct request/response mapping,
  a JSON blob goes in as a param, and a JSON blob comes out as the return.
  The Realm is free to create/load/update/delete `AuthSessions` inside the
  function. This allows better control over when new sessions are made (or
  whether to return an existing session).

### DIFF
--- a/src/github.com/matrix-org/go-neb/api.go
+++ b/src/github.com/matrix-org/go-neb/api.go
@@ -17,11 +17,11 @@ func (*heartbeatHandler) OnIncomingRequest(req *http.Request) (interface{}, *err
 	return &struct{}{}, nil
 }
 
-type configureAuthSessionHandler struct {
+type requestAuthSessionHandler struct {
 	db *database.ServiceDB
 }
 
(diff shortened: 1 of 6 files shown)


NegativeMjark:
LGTM

PR #10, opened 2016-08-05
Perform Github OAuth and store resulting tokens

Kegsay:
- Add `realmRedirectHandler` to handle redirect requests
- Add `id` column to `auth_sessions` to allow `AuthRealms` to key off a string other than the user ID.
- Implement `GithubRealm` fully.

### DIFF
--- a/src/github.com/matrix-org/go-neb/database/schema.go
+++ b/src/github.com/matrix-org/go-neb/database/schema.go
@@ -45,12 +45,14 @@ CREATE TABLE IF NOT EXISTS auth_realms (
 );
 
 CREATE TABLE IF NOT EXISTS auth_sessions (
+	id TEXT NOT NULL,
 	realm_id TEXT NOT NULL,
 	user_id TEXT NOT NULL,
 	session_json TEXT NOT NULL,
 	time_added_ms BIGINT NOT NULL,
 	time_updated_ms BIGINT NOT NULL,
-	UNIQUE(realm_id, user_id)
+	UNIQUE(realm_id, user_id),
+	UNIQUE(realm_id, id)
 );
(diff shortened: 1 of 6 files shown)


src/github.com/matrix-org/go-neb/database/schema.go
```diff
@@ -45,12 +45,14 @@ CREATE TABLE IF NOT EXISTS auth_realms (
 );
 
 CREATE TABLE IF NOT EXISTS auth_sessions (
+	id TEXT NOT NULL,
```
NegativeMjark: Can we call it session_id? It makes it clearer whats going on when constructing silly SQL.

src/github.com/matrix-org/go-neb/realms/github/github.go
```diff
@@ -60,13 +77,85 @@ func (r *githubRealm) RequestAuthSession(userID string, req json.RawMessage) int
...
+		return
+	}
+	// load the session (we keyed off the state param)
+	session, err := database.GetServiceDB().LoadAuthSessionByID(r.ID(), state)
+	if err != nil {
+		// most likely cause
+		failWith(logger, w, 400, "Provided ?state= param is not recognised.", err)
+		return
+	}
+	ghSession := session.(*githubSession)
```
NegativeMjark: It's probably unnecessary, but it might be polite to use the form of type assertion that doesn't panic.

Kegsay: done

src/github.com/matrix-org/go-neb/database/schema.go

NegativeMjark:
LGTM

PR #11, opened 2016-08-05
Make GithubService construct authenticated github clients

Kegsay:
- Add `Register()` function to Services.
- Include the `userID` in `Expand()` plugin calls.
- Pull out access tokens and use them to construct Github Clients in `GithubService`.


NegativeMjark:
LGTM

PR #13, opened 2016-08-08
Create webhooks in the Github Service

Kegsay:
- Added a new `Service` lifecycle hook: `PostRegister(oldService)`. This exists so the Github service can remove old webhooks.
- Changed config format of Github service to support webhooks.
- Listen for `ping` events from Github when webhooks are made, and reply with a pong (all that matters is a 200 status code).

The next PR will do some tidying up, notably:
 - Expose a way for services to get at their webhook URL without every service having to manually store/parse it.
 - Error-wrap SQL errors in `db.go` so you can see which statement failed (rather than unhelpful "unique key violation" errors which don't tell you what it was trying to do).

### DIFF
--- a/src/github.com/matrix-org/go-neb/services/github/github.go
+++ b/src/github.com/matrix-org/go-neb/services/github/github.go
@@ -201,6 +201,12 @@ func (s *githubService) PostRegister(oldService types.Service) {
 		return
 	}
 
+	// TODO: We should be adding webhooks in Register() then removing old hooks in PostRegister()
+	//
+	// By doing both operations in PostRegister(), if some of the requests fail we can end up in
+	// an inconsistent state. It is a lot simpler and easy to reason about this way though, so
+	// for now it will do.
+
 	// remove any existing webhooks this service created on the user's behalf
 	modifyWebhooks(old, cli, true)
 
--- a/src/github.com/matrix-org/go-neb/services/github/github.go
+++ b/src/github.com/matrix-org/go-neb/services/github/github.go
@@ -26,7 +26,11 @@ type githubService struct {
 	BotUserID    string
 	GithubUserID string
 	RealmID      string
-	WebhookRooms map[string][]string // room_id => ["push","issue","pull_request"]
+	Rooms        map[string]struct { // room_id => {}
+		OwnerRepo map[string]struct { // owner/repo => { events: ["push","issue","pull_request"] }
+			Events []string
+		}
+	}
 }
 
 func (s *githubService) ServiceUserID() string { return s.BotUserID }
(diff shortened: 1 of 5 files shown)


src/github.com/matrix-org/go-neb/services/github/github.go
```diff
@@ -167,9 +179,90 @@ func (s *githubService) Register() error {
...
+	if !ok {
+		log.Error("PostRegister: Provided old service is not of type GithubService")
+		return
+	}
+
+	// remove any existing webhooks this service created on the user's behalf
+	modifyWebhooks(old, cli, true)
+
+	// make new webhooks according to service config
+	modifyWebhooks(s, cli, false)
```
NegativeMjark: Adding the new hooks probably happen before persisting the config.

src/github.com/matrix-org/go-neb/services/github/github.go
```diff
@@ -167,9 +179,90 @@ func (s *githubService) Register() error {
...
+		return
+	}
+	old, ok := oldService.(*githubService)
+	if !ok {
+		log.Error("PostRegister: Provided old service is not of type GithubService")
+		return
+	}
+
+	// remove any existing webhooks this service created on the user's behalf
+	modifyWebhooks(old, cli, true)
```
NegativeMjark: Should probably only remove the hooks that aren't valid anymore?

Kegsay: I'm effectively `rm -rfing` everything and then starting anew, which is a lot simpler than trying to diff the two sets of configs.

src/github.com/matrix-org/go-neb/services/github/github.go
Kegsay: I need to delete old hooks before adding new hooks else I might delete a hook I just added. I can only delete them in `PostRegister`. That's why I add hooks here rather than in `Register()`.

src/github.com/matrix-org/go-neb/services/github/github.go
NegativeMjark: Okay. I guess it's not the end of the world. But I'd prefer that we failed in a way that the service continues to work where possible.

NegativeMjark: I'm happy enough to compromise to get this thing out the door, but I think a few comments explaining what the failure mode and what the user can do to fix it might be an idea.

Kegsay: /me nods

Right you are. Will explain the failure mode and tag it with a TODO.

src/github.com/matrix-org/go-neb/services/github/webhook/webhook.go
```diff
@@ -53,6 +53,10 @@ func OnReceiveRequest(r *http.Request, secretToken string) (string, *github.Repo
 		"signature":  signatureSHA1,
 	}).Print("Received Github event")
 
+	if eventType == "ping" {
```
NegativeMjark: Maybe add a comment explaining the ping/pong stuff that github does.

NegativeMjark:
Otherwise LGTM

PR #14, opened 2016-08-09
Pass in the webhook endpoint URL for each service

Kegsay:
Get the base URL from an env var 'BASE_URL' and pass it through to CreateService.


NegativeMjark:
LGTM

PR #16, opened 2016-08-10
Allow JIRA auth realms to be configured

Kegsay:
This sets up the "application linking" part of JIRA. Should be self-explanatory with the README.


src/github.com/matrix-org/go-neb/realms/jira/urls/urls_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"testing"
+)
+
+var urltests = []struct {
+	in      string
+	outKey  string
+	outBase string
+	outRaw  string
```
NegativeMjark: I found it slightly confusing that these are given in a different order than in the JIRAURL struct definition.

NegativeMjark:
Otherwise LGTM

NegativeMjark:
LGTM

PR #20, opened 2016-08-11
Implement processing of !jira create commands

Kegsay:
This involves looking up JIRA realms and then auth sessions for the user.
Does not handle starter links yet.

### DIFF
--- a/src/github.com/matrix-org/go-neb/database/schema.go
+++ b/src/github.com/matrix-org/go-neb/database/schema.go
@@ -262,6 +262,35 @@ func selectRealmTxn(txn *sql.Tx, realmID string) (types.AuthRealm, error) {
 	return realm, nil
 }
 
+const selectRealmsByTypeSQL = `
+SELECT realm_id, realm_json FROM auth_realms WHERE realm_type = $1
+`
+
+func selectRealmsByTypeTxn(txn *sql.Tx, realmType string) (realms []types.AuthRealm, err error) {
+	rows, err := txn.Query(selectRealmsByTypeSQL, realmType)
+	if err != nil {
+		return
+	}
+	defer rows.Close()
+	for rows.Next() {
+		var realmID string
+		var realmJSON []byte
+		if err = rows.Scan(&realmID, &realmJSON); err != nil {
+			return
+		}
+		realm := types.CreateAuthRealm(realmID, realmType)
+		if realm == nil {
+			err = fmt.Errorf("Cannot create realm %s of type %s", realmID, realmType)
+			return
+		}
+		if err = json.Unmarshal(realmJSON, realm); err != nil {
+			return
+		}
+		realms = append(realms, realm)
+	}
+	return
+}
+
 const updateRealmSQL = `
 UPDATE auth_realms SET realm_type=$1, realm_json=$2, time_updated_ms=$3
 	WHERE realm_id=$4
--- a/src/github.com/matrix-org/go-neb/database/schema.go
+++ b/src/github.com/matrix-org/go-neb/database/schema.go
@@ -263,7 +263,7 @@ func selectRealmTxn(txn *sql.Tx, realmID string) (types.AuthRealm, error) {
 }
 
 const selectRealmsByTypeSQL = `
-SELECT realm_id, realm_json FROM auth_realms WHERE realm_type = $1
+SELECT realm_id, realm_json FROM auth_realms WHERE realm_type = $1 ORDER BY realm_id
 `
 
 func selectRealmsByTypeTxn(txn *sql.Tx, realmType string) (realms []types.AuthRealm, err error) {
--- a/src/github.com/matrix-org/go-neb/realms/jira/jira.go
+++ b/src/github.com/matrix-org/go-neb/realms/jira/jira.go
@@ -238,15 +238,13 @@ func (r *JIRARealm) ProjectKeyExists(userID, projectKey string) (bool, error) {
 // JIRAClient returns an authenticated jira.Client for the given userID. Returns an unauthenticated
 // client if allowUnauth is true and no authenticated session is found, else returns an error.
 func (r *JIRARealm) JIRAClient(userID string, allowUnauth bool) (*jira.Client, error) {
-	var cli *jira.Client
 	// Check if user has an auth session.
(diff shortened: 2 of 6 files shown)


src/github.com/matrix-org/go-neb/services/jira/jira.go
```diff
@@ -0,0 +1,168 @@
...
+							// FIXME: This may vary depending on the JIRA install!
+							Type: jira.IssueType{
+								Name: "Bug",
+							},
+						},
+					}
+					cli, err := r.JIRAClient(userID, false)
+					i, res, err := cli.Issue.Create(&iss)
+					if err != nil {
+						log.WithError(err).Print("Failed to create issue")
```
NegativeMjark: Might want to include the project, auth realm, and user in the logging to aid debugging here?

src/github.com/matrix-org/go-neb/services/jira/jira.go
```diff
@@ -0,0 +1,168 @@
...
+		}
+
+		_, err := database.GetServiceDB().LoadAuthSessionByUser(r.ID(), userID)
+		if err != nil {
+			if err == sql.ErrNoRows {
+				// not authed with this JIRA realm; add to back of queue.
+				queue = append(queue, jrealm)
+			} else {
+				logger.WithError(err).WithField("realm_id", r.ID()).Print(
+					"Failed to load auth sessions for user")
```
NegativeMjark: I have a moderate aversion to leaving the closing ``)`` dangling on the end of the line. I think it makes it harder to work out where the expression ends.

src/github.com/matrix-org/go-neb/services/jira/jira.go
```diff
@@ -0,0 +1,168 @@
...
+			}
+			continue // this may not have been the match anyway so don't give up!
+		}
+		// authed with this JIRA realm; add to front of queue
+		queue = append([]*realms.JIRARealm{jrealm}, queue...)
```
NegativeMjark: Would it make sense to keep two queues here and to merge them together at the end?
That way the realms will be added in the same order for authed and non authed realms.

src/github.com/matrix-org/go-neb/database/schema.go
```diff
@@ -262,6 +262,35 @@ func selectRealmTxn(txn *sql.Tx, realmID string) (types.AuthRealm, error) {
 	return realm, nil
 }
 
+const selectRealmsByTypeSQL = `
+SELECT realm_id, realm_json FROM auth_realms WHERE realm_type = $1
```
NegativeMjark: Would it make sense to throw an order-by into this query so that the results come out in a consistent order?

Kegsay: The code doesn't rely on order currently. The API does not specify an order. If we need an order for whatever reason, then we can chuck in an `ORDER BY` I think.

src/github.com/matrix-org/go-neb/realms/jira/jira.go
```diff
@@ -180,39 +189,96 @@ func (r *jiraRealm) OnReceiveRedirect(w http.ResponseWriter, req *http.Request)
...
+func (r *JIRARealm) JIRAClient(userID string, allowUnauth bool) (*jira.Client, error) {
+	var cli *jira.Client
+	// Check if user has an auth session.
+	session, err := database.GetServiceDB().LoadAuthSessionByUser(r.id, userID)
+	if err != nil {
+		if err == sql.ErrNoRows {
+			if allowUnauth {
+				// make an unauthenticated client
+				cli, err = jira.NewClient(nil, r.JIRAEndpoint)
+				return cli, err
```
NegativeMjark: Could write this as ``return jira.NewClient(nil, r.JIRAEndpoint)``

src/github.com/matrix-org/go-neb/services/jira/jira.go
Kegsay: I also don't like it dangling at the end of the line but it won't compile otherwise :/

Kegsay: Oh you have to put a trailing comma, then you can

src/github.com/matrix-org/go-neb/database/schema.go
NegativeMjark: I think the behaviour of ``projectToRealm`` depends on the order that the realms are returned in?

NegativeMjark:
Hmm, I wonder if the list of JIRA realms to search for projects in should be part of the service config, rather than searching all of the realms configured in go-neb?

src/github.com/matrix-org/go-neb/database/schema.go
Kegsay: Aha, you're right. Okay, will ORDER BY.

Kegsay:
I think that only matters if you want to **not** search some realms. When would we ever want to not just search all the JIRA realms for a project match?

src/github.com/matrix-org/go-neb/services/jira/jira.go
Kegsay: done

src/github.com/matrix-org/go-neb/services/jira/jira.go
Kegsay: done

src/github.com/matrix-org/go-neb/services/jira/jira.go
Kegsay: done

src/github.com/matrix-org/go-neb/database/schema.go

src/github.com/matrix-org/go-neb/realms/jira/jira.go
Kegsay: done

Kegsay:
PTAL @NegativeMjark

NegativeMjark:
I guess it depends on whether you want to add a "JIRA" integration that attempts to integrate with all JIRAs in existence on the public internet. Or whether you want to add an integration for a specific JIRA instance.

Kegsay:
I guess so. I don't know the answer to that question though.

NegativeMjark:
LGTM

PR #21, opened 2016-08-12
Always create realms/services consistently

Kegsay:
- Always unmarshal JSON into realms/services
- Always `Init()` realms.


NegativeMjark:
LGTM

PR #22, opened 2016-08-12
Expand JIRA issues as they are mentioned

Kegsay:
Based off https://github.com/matrix-org/go-neb/pull/21

### DIFF
--- a/src/github.com/matrix-org/go-neb/services/jira/jira.go
+++ b/src/github.com/matrix-org/go-neb/services/jira/jira.go
@@ -129,21 +129,29 @@ func (s *jiraService) expandIssue(roomID, userID, issueKey string) interface{} {
 		return nil
 	}
 
+	r, err := database.GetServiceDB().LoadAuthRealm(s.Rooms[roomID].RealmID)
+	if err != nil {
+		logger.WithFields(log.Fields{
+			"realm_id":   s.Rooms[roomID].RealmID,
+			log.ErrorKey: err,
+		}).Print("Failed to load realm")
+		return nil
+	}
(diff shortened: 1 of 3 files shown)


src/github.com/matrix-org/go-neb/services/jira/jira.go
```diff
@@ -174,6 +261,20 @@ func (s *jiraService) projectToRealm(userID, pkey string) (*realms.JIRARealm, er
...
+	//   "Flibble Wibble [P1, In Progress]"
+	status := html.EscapeString(issue.Fields.Status.Name)
+	if issue.Fields.Resolution != nil {
+		status = fmt.Sprintf("%s (%s)",
+			status, html.EscapeString(issue.Fields.Resolution.Name))
+	}
+	return fmt.Sprintf("%s [%s, %s]",
+		html.EscapeString(issue.Fields.Summary),
+		html.EscapeString(issue.Fields.Priority.Name),
+		status)
```
NegativeMjark: I'm not sure about the dangling ')'

Kegsay: Bah thought I got them all, thanks!

NegativeMjark:
Apart from the dangling ')' LGTM.

At somepoint we might want to think about the utility of the error messages. I suspect that the error messages will need to include enough information for the end user to correct their mistakes and typos and I don't think that the current messages give enough information for them to do so.

But I'm not sure how we want to go about doing that, So I'm happy to leave them as they are for now.

Kegsay:
Agreed, there's been places where I've wanted to return more human friendly errors but not had the right error bubbling to do it.

PR #23, opened 2016-08-12
Implement JIRA webhook creation

Kegsay:
Also start including unauthorised webhook creation semantics, though not
finished yet. Change the config format of `JIRAService` to make more sense
and allow multiple realms per room.

Still left to do:
 - Handle incoming webhook requests from JIRA
 - Check if there is already a webhook set up for the given JIRA endpoint so unauthorised users can create services.

### DIFF
--- a/src/github.com/matrix-org/go-neb/services/jira/jira.go
+++ b/src/github.com/matrix-org/go-neb/services/jira/jira.go
@@ -10,6 +10,7 @@ import (
 	"github.com/matrix-org/go-neb/matrix"
 	"github.com/matrix-org/go-neb/plugin"
 	"github.com/matrix-org/go-neb/realms/jira"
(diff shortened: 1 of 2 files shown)


src/github.com/matrix-org/go-neb/services/jira/jira.go
```diff
@@ -125,21 +149,36 @@ func (s *jiraService) expandIssue(roomID, userID, issueKey string) interface{} {
...
+				break
+			}
+		}
+		if realmID != "" {
+			break
+		}
+	}
+	if realmID == "" {
 		return nil
 	}
```
NegativeMjark: Maybe the above loop should be a separate function?

Kegsay: done

NegativeMjark:
LGTM

PR #24, opened 2016-08-12
Handle incoming JIRA webhook events

Kegsay:
Also support unauthenticated users piggy-backing off existing webhooks.


NegativeMjark:
LGTM

PR #25, opened 2016-08-15
Add endpoint to GET a Service config

Kegsay:


src/github.com/matrix-org/go-neb/api.go
```diff
@@ -228,3 +229,37 @@ func (s *configureServiceHandler) OnIncomingRequest(req *http.Request) (interfac
 		NewConfig types.Service
 	}{body.ID, body.Type, oldService, service}, nil
 }
+
+type getServiceHandler struct {
+	db *database.ServiceDB
+}
+
+func (h *getServiceHandler) OnIncomingRequest(req *http.Request) (interface{}, *errors.HTTPError) {
+	if req.Method != "GET" {
```
NegativeMjark: GET with a body is probably a bad idea.

NegativeMjark: The most obvious objection being that the go http client doesn't support without building the request by hand.

Kegsay: Oops, my bad.

Kegsay: POST now

NegativeMjark:
LGTM

PR #26, opened 2016-08-15
Add endpoint for GETing AuthSession information for a user/realm tuple

Kegsay:
See https://github.com/matrix-org/go-neb/pull/25 first as this PR is based off that one.

Add a new `AuthSession` function `Authenticated()` which returns `true` if the
user has completed the auth process. This allows the caller to distinguish
between:
 - Never done any auth (404s)
 - In the process of doing auth (`Authenticated == false`)
 - Finished doing auth (`Authenticated == true`)

This does **not** yet list *possible* repos for the session.


NegativeMjark:
LGTM.

PR #27, opened 2016-08-15
Add StarterLink as a field to the Github Service config

Kegsay:
Use it verbatim in the responses to !commands when the caller has not authed
with Github.

~~It feels a little wrong that this is a property of the service rather than the realm but \*shrug\*~~ Going to do this by realm because that's what naturally falls out with JIRA and I want starter links to consistently be stored in the same entity type.


NegativeMjark:
LGTM, maybe document what a starter link is?

PR #29, opened 2016-08-15
List possible GH repos on /getSession

Kegsay:
Add an `Info()` method to pull out info about an auth session.

### DIFF
--- a/src/github.com/matrix-org/go-neb/api.go
+++ b/src/github.com/matrix-org/go-neb/api.go
@@ -295,6 +295,6 @@ func (h *getSessionHandler) OnIncomingRequest(req *http.Request) (interface{}, *
 	return &struct {
 		ID            string
 		Authenticated bool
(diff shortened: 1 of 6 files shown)


NegativeMjark:
LGTM

src/github.com/matrix-org/go-neb/api.go
```diff
@@ -295,6 +295,6 @@ func (h *getSessionHandler) OnIncomingRequest(req *http.Request) (interface{}, *
 	return &struct {
 		ID            string
 		Authenticated bool
-		Session       types.AuthSession
-	}{session.ID(), session.Authenticated(), session}, nil
+		Session       interface{}
```
NegativeMjark: Actually is ``Session`` the right name for this?

Kegsay: Renamed to `Info` as per irl.

Issue #18, opened 2016-08-11
plugin.Expansion Expand() should include matching groups


Kegsay:
I find myself having to run the regexp twice because the `Expand()` callback does not give me any captured groups (just the `matchingText` string). I run it once so the callback fires, then run it again to actually extract the useful bits.

I think it is relatively rare to have a case where there is **no** processing to be done on the resulting match, so I wonder if we should replace the `string` with a `[]string` of matched groups. All these cases I can think of require more processing which a matching group can extract:
 - `RFC 1234` => `1234`
 - `matrix-org/go-neb#1234` => `[matrix-org/go-neb, 1234]`
 - `SYN-1234` => `[SYN, 1234]`

PR #32, opened 2016-08-16
Change Expansion.Expand to include matching groups rather than just the match

Kegsay:
Much more useful like this.

Fixes https://github.com/matrix-org/go-neb/issues/18


NegativeMjark:
LGTM

PR #35, opened 2016-08-17
Base64 redirect and webhook paths

Kegsay:
URL-safe base64 encode the service ID in the webhook path and the realm ID in the redirect path.


NegativeMjark:
LGTM

PR #36, opened 2016-08-19
Change how Services are notified for incoming !commands and expansions

Kegsay:
Previously, we would notify `Services` based on matching the `room_id` of the
event with a list of `RoomIDs()` which the service returned.

Now we notify `Services` based on matching the `user_id` of the client listening
for events. This means that the service will receive more events because there
isn't a filter on a set of room IDs.

This is required in order to implement "auto-join on invite" semantics for
Services, as the room ID is not known at that point in time.

Also add flags for `HandleCommands` and `HandleExpansions` to control whether the service should handle them when they receive them.

### DIFF
--- a/src/github.com/matrix-org/go-neb/database/schema.go
+++ b/src/github.com/matrix-org/go-neb/database/schema.go
@@ -18,6 +18,7 @@ CREATE TABLE IF NOT EXISTS services (
 	time_updated_ms BIGINT NOT NULL,
 	UNIQUE(service_id)
 );
+CREATE UNIQUE INDEX IF NOT EXISTS service_id_and_user_idx ON services(service_user_id, service_id);
 
 CREATE TABLE IF NOT EXISTS matrix_clients (
 	user_id TEXT NOT NULL,
--- a/src/github.com/matrix-org/go-neb/database/schema.go
+++ b/src/github.com/matrix-org/go-neb/database/schema.go
@@ -12,18 +12,11 @@ const schemaSQL = `
 CREATE TABLE IF NOT EXISTS services (
 	service_id TEXT NOT NULL,
 	service_type TEXT NOT NULL,
+	service_user_id TEXT NOT NULL,
 	service_json TEXT NOT NULL,
 	time_added_ms BIGINT NOT NULL,
 	time_updated_ms BIGINT NOT NULL,
 	UNIQUE(service_id)
-);
-
-CREATE TABLE IF NOT EXISTS rooms_to_services (
-	service_user_id TEXT NOT NULL,
-	room_id TEXT NOT NULL,
-	service_id TEXT NOT NULL,
-	time_added_ms BIGINT NOT NULL,
-	UNIQUE(service_user_id, room_id, service_id)
 );
 
 CREATE TABLE IF NOT EXISTS matrix_clients (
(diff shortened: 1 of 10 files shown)


src/github.com/matrix-org/go-neb/database/schema.go
```diff
@@ -12,20 +12,13 @@ const schemaSQL = `
 CREATE TABLE IF NOT EXISTS services (
 	service_id TEXT NOT NULL,
 	service_type TEXT NOT NULL,
+	service_user_id TEXT NOT NULL,
 	service_json TEXT NOT NULL,
 	time_added_ms BIGINT NOT NULL,
 	time_updated_ms BIGINT NOT NULL,
 	UNIQUE(service_id)
 );
```
NegativeMjark: Need to add an index on service_user_id, service_id

src/github.com/matrix-org/go-neb/services/github/github.go
```diff
@@ -24,30 +24,27 @@ var ownerRepoIssueRegex = regexp.MustCompile("([A-z0-9-_]+)/([A-z0-9-_]+)#([0-9]
 
 type githubService struct {
 	id                 string
+	serviceUserID      string
```
NegativeMjark: Maybe make a 
```go
type BaseService struct {
    id string
    serviceUserID string
}
```

src/github.com/matrix-org/go-neb/database/schema.go
Kegsay: Why? I don't see anything wrong with 1 user ID mapping to >1 Service?

src/github.com/matrix-org/go-neb/services/github/github.go
```diff
@@ -24,30 +24,27 @@ var ownerRepoIssueRegex = regexp.MustCompile("([A-z0-9-_]+)/([A-z0-9-_]+)#([0-9]
...
 	webhookEndpointURL string
-	BotUserID          string
 	ClientUserID       string
 	RealmID            string
 	SecretToken        string
+	HandleCommands     bool
+	HandleExpansions   bool
```
NegativeMjark: Should this be part of the Service interface so that they can be indexed in the database?

Kegsay: Not planning to do that right now given that commands/expansions are already efficient (simple SELECT based on the user_id of the client doing the `/sync`). Maybe in the future if we need to.

src/github.com/matrix-org/go-neb/services/github/github.go
Kegsay: Something like that sounds good to me. Will do.

NegativeMjark:
Otherwise LGTM

src/github.com/matrix-org/go-neb/database/schema.go
NegativeMjark: Full table scans make me sad.

Kegsay: oh woops

src/github.com/matrix-org/go-neb/services/github/github.go
Kegsay: https://github.com/matrix-org/go-neb/issues/37

PR #38, opened 2016-08-19
Start client sync streams on startup

Kegsay:
- If the `Sync` flag is set, start syncing.
- If the `AutoJoinRooms` flag is set, automatically join rooms when invited.


NegativeMjark:
LGTM

PR #39, opened 2016-08-19
Client ID is optional when creating Github services

Kegsay:
We only use it for making webhooks (to determine which GH account to load up).
We need it to be optional for the global GH bot to work (which has no account
and uses the account of whoever issued the !command to do the work).

### DIFF
--- a/src/github.com/matrix-org/go-neb/services/github/github.go
+++ b/src/github.com/matrix-org/go-neb/services/github/github.go
@@ -224,7 +224,11 @@ func (s *githubService) PostRegister(oldService types.Service) {
 	// PostRegister handles creating/destroying webhooks, which is only valid if this service
 	// is configured on behalf of a client.
 	if s.ClientUserID == "" {
-		// TODO: We should log an error here if the service has any webhook config in Rooms
+		if len(s.Rooms) != 0 {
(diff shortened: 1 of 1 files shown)


src/github.com/matrix-org/go-neb/services/github/github.go
```diff
@@ -219,6 +221,12 @@ func (s *githubService) Register() error {
 }
 
 func (s *githubService) PostRegister(oldService types.Service) {
+	// PostRegister handles creating/destroying webhooks, which is only valid if this service
+	// is configured on behalf of a client.
+	if s.ClientUserID == "" {
+		return
```
NegativeMjark: It'd be nice if it error'd when trying to add webhooks without a ClientUserID.

NegativeMjark:
Other than the lack of errors when a user fails to supply a client id it LGTM.

src/github.com/matrix-org/go-neb/services/github/github.go
Kegsay: That makes no sense. `PostRegister` is called regardless of whether any webhooks are set.

NegativeMjark:
LGTM

PR #43, opened 2016-08-22
Create delta webhooks in GithubService.Register()

Kegsay:
Remove `PostRegister()` as it isn't used for anything anymore. Add TODO marker
for guarding against race conditions, which will arrive in the next PR.

### DIFF
--- a/src/github.com/matrix-org/go-neb/util/util.go
+++ b/src/github.com/matrix-org/go-neb/util/util.go
@@ -1,0 +1,35 @@
+package util
+
+import (
+	"sort"
+)
+
+// Difference returns the elements that are only in the first list and
+// the elements that are only in the second. As a side-effect this sorts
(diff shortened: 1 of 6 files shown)


src/github.com/matrix-org/go-neb/util/util.go
```diff
@@ -0,0 +1,35 @@
+package util
+
+import (
+	"sort"
+)
+
+// Difference returns the elements that are only in the first list and
+// the elements that are only in the second. As a side-effect this sorts
+// the input lists in-place.
+func Difference(a, b []string) (onlyA, onlyB []string) {
```
NegativeMjark: Probably not worth splitting this out to a separate file until there are two places it gets called.

NegativeMjark:
Other than maybe splitting out the Difference function prematurely LGTM.

PR #44, opened 2016-08-22
Mutex-guard services with the same ID being modified concurrently

Kegsay:
Also treat 422 "already exists" errors as success.

### DIFF
--- a/src/github.com/matrix-org/go-neb/api.go
+++ b/src/github.com/matrix-org/go-neb/api.go
@@ -11,6 +11,7 @@ import (
 	"github.com/matrix-org/go-neb/types"
 	"net/http"
 	"strings"
+	"sync"
 )
 
 type heartbeatHandler struct{}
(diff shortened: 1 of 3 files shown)


src/github.com/matrix-org/go-neb/api.go
```diff
@@ -202,8 +203,29 @@ func (s *configureClientHandler) OnIncomingRequest(req *http.Request) (interface
...
+	}
+}
+
+func (s *configureServiceHandler) getMutexForServiceID(serviceID string) *sync.Mutex {
+	s.mapMutex.Lock()
+	defer s.mapMutex.Unlock()
+	m := s.mutexByServiceID[serviceID]
+	if m == nil {
+		m = &sync.Mutex{}
+		s.mutexByServiceID[serviceID] = m
```
NegativeMjark: Do we ever remove the mutexes from mutexByServiceID?

Kegsay: No we do not, but I can't make the mutex a property of the Service since we constantly create/destroy them *per request*.

NegativeMjark: I'd prefer that you did something sensible like a striped mutex, (or just used a single mutex for all services here.). I'm not really happy with the idea of creating N mutexes where N == number of services.

Kegsay: Why aren't you happy with 1 mutex per service?

NegativeMjark: Because you never remove them from the map? so the map of mutexes grows without limit?

Kegsay: I don't think that will be a serious problem, and if memory leaking 10s of bytes is seriously an issue, we can optimise that later down the line.

NegativeMjark:
LGTM

PR #46, opened 2016-08-22
Join webhook rooms on Register()

Kegsay:
Fixes https://github.com/matrix-org/go-neb/issues/41


NegativeMjark:
LGTM once #44 and #45 land.

PR #47, opened 2016-08-24
Support bot options and implement github bot options for default repos

Kegsay:
- Add a new `bot_options` table
- Allow github expansions/commands to omit the owner/repo and use a default if set


NegativeMjark:
LGTM

PR #52, opened 2016-08-30
Split up github service into 2 services

Kegsay:
One for commands/expansions, the other for webhooks. Call the type 'github-webhook'.


NegativeMjark:
LGTM

PR #56, opened 2016-09-02
Set the desired DisplayName for a client when it is being configured

Kegsay:
Also, only start a sync stream if we have been instructed to do so (`github-webhook` services typically don't, whereas `github` service does).


NegativeMjark:
LGTM

PR #57, opened 2016-09-02
Allow github webhooks to be explicitly removed

Kegsay:
Removing a repo entirely from a service will now remove the webhook in the
`PostRegister()` function. This function is still within the `configureService`
critical section to prevent another request for the same service racing with
the cleanup process.

### DIFF
--- a/src/github.com/matrix-org/go-neb/services/github/github_webhook.go
+++ b/src/github.com/matrix-org/go-neb/services/github/github_webhook.go
@@ -105,14 +105,9 @@ func (s *githubWebhookService) Register(oldService types.Service, client *matrix
 	if s.RealmID == "" || s.ClientUserID == "" {
 		return fmt.Errorf("RealmID and ClientUserID is required")
 	}
-	// check realm exists
-	realm, err := database.GetServiceDB().LoadAuthRealm(s.RealmID)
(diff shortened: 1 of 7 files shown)


src/github.com/matrix-org/go-neb/services/github/github_webhook.go
```diff
@@ -163,6 +161,39 @@ func (s *githubWebhookService) Register(oldService types.Service, client *matrix
...
+
+	// Register() handled adding the new repos, we just want to clean up after ourselves
+	_, removedRepos := util.Difference(newRepos, oldRepos)
+	for _, r := range removedRepos {
+		segs := strings.Split(r, "/")
+		if err := s.deleteHook(segs[0], segs[1]); err != nil {
+			log.WithFields(log.Fields{
+				log.ErrorKey: err,
+				"repo":       r,
+			}).Error("Failed to remove webhook")
```
NegativeMjark: Maybe warn here?

NegativeMjark:
LGTM

src/github.com/matrix-org/go-neb/services/github/github_webhook.go
Kegsay: done

PR #60, opened 2016-09-06
Persist next_batch tokens

Kegsay:
- Indirect this through a `NextBatchStorer` interface type to get around circular dependencies
which are formed if you try to inline it in the `Sync()` while loop.

- Fork out event handlers to separate functions in `Clients` to make gocyclo happy.

- Add `SELECT` and `UPDATE` queries for `next_batch`.

- Set the filter limit to `50` to actually get >1 timeline event.

- Check if the `next_batch` was originally the empty string and if so, do not pass the results to the `Worker` as this is the first sync evar (subsequent syncs after restarts use an existing token in the database).

### DIFF
--- a/src/github.com/matrix-org/go-neb/matrix/matrix.go
+++ b/src/github.com/matrix-org/go-neb/matrix/matrix.go
@@ -47,17 +47,17 @@ func (s noopNextBatchStore) Load(userID string) string     { return "" }
 
 // Client represents a Matrix client.
 type Client struct {
-	HomeserverURL *url.URL
-	Prefix        string
-	UserID        string
-	AccessToken   string
-	Rooms         map[string]*Room
-	Worker        *Worker
-	syncingMutex  sync.Mutex
-	syncingID     uint32 // Identifies the current Sync. Only one Sync can be active at any given time.
-	httpClient    *http.Client
-	filterID      string
-	NextBatch     NextBatchStorer
+	HomeserverURL   *url.URL
+	Prefix          string
+	UserID          string
+	AccessToken     string
+	Rooms           map[string]*Room
+	Worker          *Worker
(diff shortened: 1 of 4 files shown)


src/github.com/matrix-org/go-neb/matrix/matrix.go
```diff
@@ -366,6 +390,7 @@ func NewClient(homeserverURL *url.URL, accessToken string, userID string) *Clien
 		Prefix:        "/_matrix/client/r0",
 	}
 	cli.Worker = newWorker(&cli)
+	cli.NextBatch = noopNextBatchStore{}
```
NegativeMjark: This is cute, but maybe could do with a comment?

src/github.com/matrix-org/go-neb/matrix/matrix.go
```diff
@@ -43,6 +57,7 @@ type Client struct {
 	syncingID     uint32 // Identifies the current Sync. Only one Sync can be active at any given time.
 	httpClient    *http.Client
 	filterID      string
+	NextBatch     NextBatchStorer
```
NegativeMjark: Maybe call the member ``NextBatchStorer``?

NegativeMjark:
LGTM apart from needing a few comments to explain the sneakiness going on with the noopNextBatchStore.

src/github.com/matrix-org/go-neb/matrix/matrix.go
Kegsay: done

src/github.com/matrix-org/go-neb/matrix/matrix.go
Kegsay: done

PR #61, opened 2016-09-07
Do not process /sync responses for rooms the bot has just joined

Kegsay:
Else we can process !commands multiple times and/or process !commands from before
we were in the room (if `history_visibility` allows it).

### DIFF
--- a/src/github.com/matrix-org/go-neb/matrix/matrix.go
+++ b/src/github.com/matrix-org/go-neb/matrix/matrix.go
@@ -280,6 +280,7 @@ func (cli *Client) shouldProcessResponse(tokenOnSync string, syncResponse *syncH
 	//
 	// Work around this by inspecting each room's timeline and seeing if an m.room.member event for us
 	// exists and is "join" and then discard processing that room entirely if so.
+	// TODO: We probably want to process the !commands from after the last join event in the timeline.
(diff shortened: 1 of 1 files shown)


src/github.com/matrix-org/go-neb/matrix/matrix.go
```diff
@@ -263,12 +261,54 @@ func (cli *Client) Sync() {
...
+func (cli *Client) shouldProcessResponse(tokenOnSync string, syncResponse *syncHTTPResponse) bool {
+	if tokenOnSync == "" {
+		return false
+	}
+	// This is a horrible hack because /sync will return the most recent messages for a room
+	// as soon as you /join it. We do NOT want to process those events in that particular room
+	// because they may have already been processed (if you toggle the bot in/out of the room).
+	//
+	// Work around this by inspecting each room's timeline and seeing if an m.room.member event for us
+	// exists and is "join" and then discard processing that room entirely if so.
```
NegativeMjark: I wonder if you'd want to process the !commands from after the last join event in the timeline.

NegativeMjark:
I'm happy enough for this code to land since it's probably an improvement on the current situation. Might want to add a TODO marker for handling the timeline messages from after the join event.

PR #62, opened 2016-09-07
Delete webhook services when there are no webhooks left

Kegsay:
Else they will show up in the current integration list FOREVAR!


NegativeMjark:
LGTM

PR #64, opened 2016-09-08
Add /admin/removeAuthSession

Kegsay:
Allows users to nuke their OAuth credentials. Currently this just does a database `DELETE` as neither Github nor JIRA need you to hit any special `/logout` API.


NegativeMjark:
LGTM

PR #67, opened 2016-09-09
Improve !github create parsing

Kegsay:
To handle cases where `args[0]` has a `/` but isn't an `owner/repo`.


NegativeMjark:
LGTM

PR #68, opened 2016-09-10
Slack API service.

aviraldg:
This is not quite complete (in that it does not implement the format described by Slack in its entirety), but should be usable for most practical purposes. I do not have much experience with Go, so do let me know if I'm doing something obviously wrong.


src/github.com/matrix-org/go-neb/services/slackapi/message.go
```diff
@@ -0,0 +1,177 @@
...
+func getSlackMessage(req http.Request) (message slackMessage, err error) {
+	ct := req.Header.Get("Content-Type")
+	ct, _, err = mime.ParseMediaType(ct)
+
+	if ct == "application/x-www-form-urlencoded" {
+		req.ParseForm()
+		payload := req.Form.Get("payload")
+		err = json.Unmarshal([]byte(payload), &message)
+	} else if ct == "application/json" {
+		log.Info("Parsing as JSON")
```
Kegsay: Remove this please, it isn't very informative.

src/github.com/matrix-org/go-neb/services/slackapi/message.go
```diff
@@ -0,0 +1,177 @@
...
+		decoder := json.NewDecoder(req.Body)
+		err = decoder.Decode(&message)
+	} else {
+		message.Text = fmt.Sprint("**Error:** unknown Content-Type `%s`", ct)
+		log.Error(message.Text)
```
Kegsay: Why aren't we returning an error here when it clearly is?

src/github.com/matrix-org/go-neb/services/slackapi/message.go
```diff
@@ -0,0 +1,177 @@
...
+
+	return
+}
+
+func linkifyString(text string) string {
+	return linkRegex.ReplaceAllString(text, "<a href=\"$1\">$3</a>")
+}
+
+func getColor(color *string) string {
+	if color != nil {
```
Kegsay: Invert and outdent please.

src/github.com/matrix-org/go-neb/services/slackapi/message.go
```diff
@@ -0,0 +1,177 @@
...
+		// https://api.slack.com/docs/message-attachments defines these aliases
+		mappedColor, ok := map[string]string{
+			"good":    "green",
+			"warning": "yellow",
+			"danger":  "red",
+		}[*color]
+		if ok {
+			return mappedColor
+		}
+		return *color
```
Kegsay: This feels weird. If `color` is "the quick brown fox jumps over the lazy dog" we just blindly return that without a second thought. Why aren't we validating this more?

src/github.com/matrix-org/go-neb/services/slackapi/message.go
```diff
@@ -0,0 +1,177 @@
...
+		}
+
+		for _, fieldName := range attachment.MrkdwnIn {
+			var targetField, srcField *string
+
+			switch fieldName {
+			case "text":
+				srcField = &attachment.Text
+				targetField = &target.Text
+				break
```
Kegsay: Go breaks by default so this isn't necessary.

src/github.com/matrix-org/go-neb/services/slackapi/message.go
```diff
@@ -0,0 +1,177 @@
...
+		}[*color]
+		if ok {
+			return mappedColor
+		}
+		return *color
+	}
+	return "black"
+}
+
+func slackMessageToHTMLMessage(message slackMessage) (html matrix.HTMLMessage, err error) {
```
Kegsay: Named returns for functions as long as this are really hard to read. Please stick to plain old returns in this case.

src/github.com/matrix-org/go-neb/services/slackapi/slackapi.go
```diff
@@ -0,0 +1,58 @@
...
+	"net/http"
+	"strings"
+)
+
+type slackAPIService struct {
+	id                 string
+	serviceUserID      string
+	webhookEndpointURL string
+	ClientUserID       string
+	Hooks              map[string]string
```
Kegsay: Please in line comment what this maps from and to.

src/github.com/matrix-org/go-neb/services/slackapi/message.go
aviraldg: Unintentionally included.

src/github.com/matrix-org/go-neb/services/slackapi/slackapi.go
```diff
@@ -0,0 +1,58 @@
...
+func (s *slackAPIService) Register(oldService types.Service, client *matrix.Client) error {
+	return nil
+}
+
+func (s *slackAPIService) Plugin(cli *matrix.Client, roomID string) plugin.Plugin {
+	return plugin.Plugin{}
+}
+
+func (s *slackAPIService) OnReceiveWebhook(w http.ResponseWriter, req *http.Request, cli *matrix.Client) {
+	segments := strings.Split(req.URL.Path, "/")
```
Kegsay: Length checks please! `len(segments-3)` may be - ve!

src/github.com/matrix-org/go-neb/services/slackapi/slackapi.go
```diff
@@ -0,0 +1,58 @@
...
+	hookID := segments[len(segments)-2]
+	messageType := segments[len(segments)-3]
+	roomID := s.Hooks[hookID]
+
+	slackMessage, err := getSlackMessage(*req)
+	if err != nil {
+		return
+	}
+	htmlMessage, err := slackMessageToHTMLMessage(slackMessage)
```
Kegsay: Why don't we do anything with this `err`? Should probably be returning before we send messages.

src/github.com/matrix-org/go-neb/services/slackapi/message.go
aviraldg: Is there a better way to do this (ie. if a field name is in `mrkdwn_in`, convert it from Markdown) than a switch-case?

Kegsay: Nope. Go is very mechanical and doesn't really *do* any sort of metaprogramming which you can do in JS. 

As an aside, I don't think you're doing *anything* in this for loop ( `for _, fieldName := range attachment.MrkdwnIn`), you assign to src and target fields (declared anew each loop), then set target field after passing it through `blackfriday`, then uh, drop it to the floor and do nothing with the result :)

src/github.com/matrix-org/go-neb/services/slackapi/message.go
aviraldg: That would mean users on the Matrix side would see nothing (not good, and difficult to diagnose)

src/github.com/matrix-org/go-neb/services/slackapi/message.go
aviraldg: `the quick brown fox jumps over the lazy dog` is also a color :smile:

src/github.com/matrix-org/go-neb/services/slackapi/slackapi.go
aviraldg: I understand what you're trying to say, but I don't think this can happen (practically), because /services/hooks/id itself is 3 segments. Hence my question in #neb about whether this kind of URL "parsing" is sufficient.

Kegsay: You're right in that it cannot happen in its current form. However, changing the path in another file should not cause the entire process to panic and die. Without length checks that's exactly what will happen.

Kegsay:
Let me know when you want me to look at this again.

brunoqc:
@aviraldg: any progress?

aviraldg:
I'm waiting for a review from @Kegsay

brunoqc:
Oh. I thought he was waiting after you.

src/github.com/matrix-org/go-neb/services/slackapi/message.go
```diff
@@ -0,0 +1,219 @@
...
+	TextRendered template.HTML
+	Username     string            `json:"username"`
+	Channel      string            `json:"channel"`
+	Mrkdwn       *bool             `json:"mrkdwn"`
+	Attachments  []slackAttachment `json:"attachments"`
+}
+
+// We use text.template because any fields of any attachments could
+// be Markdown, so it's convenient to escape on a field-by field basis.
+// We do not do this yet, since it's assumed that clients also escape the content we send them.
```
Kegsay: > it's assumed that clients also escape the content we send them

It is, but we should be good Matrix citizens where possible!

src/github.com/matrix-org/go-neb/services/slackapi/message.go
```diff
@@ -0,0 +1,219 @@
...
+	{{- if .Pretext }}{{ or .PretextRendered .Pretext }}<br />{{ end }}
+	{{- if .Text }}{{ or .TextRendered .Text }}<br />{{ end }}
+{{- end }}
+`)
+
+var netClient = &http.Client{
+	Timeout: time.Second * 10,
+}
+
+var linkRegex, _ = regexp.Compile("<([^|]+)(\\|([^>]+))?>")
```
Kegsay: I don't know about you, but I look at this and I think "what does this do?" - Can you comment what this is doing please? E.g:
```go
// Extract hostname
// https://foobar.com/example/string =>  foobar.com 
var linkRegex, _ = regexp.Compile("<([^|]+)(\\|([^>]+))?>")
```

src/github.com/matrix-org/go-neb/services/slackapi/message.go
Kegsay: Still there.

src/github.com/matrix-org/go-neb/services/slackapi/message.go
```diff
@@ -0,0 +1,219 @@
...
+		"danger":  "red",
+	}[*color]
+	if ok {
+		return mappedColor
+	}
+	return *color
+}
+
+// fetches an image and encodes it as a data URL
+// returns nil if fetch fails
```
Kegsay: No it doesn't :angel: It returns a 0-length string. Named return args are default-initialised and `template.URL` is an alias for a `string`.

src/github.com/matrix-org/go-neb/services/slackapi/message.go
```diff
@@ -0,0 +1,219 @@
...
+func fetchAndEncodeImage(url *string) (data template.URL) {
+	if url == nil {
+		return
+	}
+
+	var resp *http.Response
+	resp, err := netClient.Get(*url)
+	if err == nil {
```
Kegsay: Invert and outdent please!

src/github.com/matrix-org/go-neb/services/slackapi/message.go
```diff
@@ -0,0 +1,219 @@
...
+		case "pretext":
+			srcField = &attachment.Pretext
+			targetField = &attachment.PretextRendered
+		case "fallback":
+			srcField = &attachment.Fallback
+			targetField = &attachment.FallbackRendered
+		}
+
+		if targetField != nil && srcField != nil {
+			log.Info(targetField)
```
Kegsay: Can you make this more informative and less spammy please.

src/github.com/matrix-org/go-neb/services/slackapi/slackapi.go
```diff
@@ -0,0 +1,73 @@
...
+func (s *slackAPIService) Plugin(cli *matrix.Client, roomID string) plugin.Plugin {
+	return plugin.Plugin{}
+}
+
+func (s *slackAPIService) OnReceiveWebhook(w http.ResponseWriter, req *http.Request, cli *matrix.Client) {
+	segments := strings.Split(req.URL.Path, "/")
+
+	if len(segments) < 2 {
+		w.WriteHeader(400)
+	}
```
Kegsay: Missing `return`?

brunoqc: You also can use http.StatusBadRequest instead of 400.

Kegsay: Indeed, though this project doesn't set any particular standard on which to use, so both are acceptable.

Kegsay:
@aviraldg - Any update here? I'd like to get this merged in at some point. Might take it over if there is no more progress.

Kegsay:
I will take this over in the next week or so.

Kegsay:
As per https://matrix.to/#/!aZkanAnWEdxcRIQkWn:matrix.org/$14781805511639206dIrhH:matrix.org - taking this over.

Kegsay:
https://github.com/matrix-org/go-neb/tree/feature-slack-api

brunoqc:
Will you update this issue when it's done or open a new one so I can subscribe to it?

Kegsay:
https://github.com/matrix-org/go-neb/pull/116 - Do you have any good way I can test whatever this is supposed to do? 0:)

brunoqc:
I'm planing to use the slack-api with prometheus's alertmanager.

Here's a ngrep output. Is that enough?

```
T 10.1.10.23:52242 -> 45.33.117.213:80 [AP]
  POST /hooks/!sMQokYMsFhPAZjMpbZ:localhost HTTP/1.1..Host: 327dfd88.ngrok.io..User-Agent: Go-http-client/1.1..Content-Length: 416..Content-Type: application/json..Accept-Encoding: g
  zip....{"channel":"test","username":"AlertManager","attachments":[{"title":"[FIRING:1] testAlert2 (test-job localhost:9091 pushgateway codelab-monitor)","title_link":"http://ubuntu
  new:9093/#/alerts?receiver=bruno","text":"","fallback":"[FIRING:1] testAlert2 (test-job localhost:9091 pushgateway codelab-monitor) | http://ubuntunew:9093/#/alerts?receiver=bruno"
  ,"color":"danger","mrkdwn_in":["fallback","pretext","text"]}]}.
```

PR #76, opened 2016-09-30
Implement polling for Services

Kegsay:
- Services can now return a `Poller` which will be polled at set intervals.
- Services now extend the `DefaultService` so no-op methods can be entirely omitted.
- There is a stub `rssService` which is not currently active.


Kegsay:
Walked through IRL with @lukebarnard1

PR #78, opened 2016-10-05
Implement feedreader service

Kegsay:
Notable improvements on this would be:
 - We currently do 1 goroutine per service. This could be bad if we have lots of these things running around.
 - We do not cache the response to RSS feeds. If we have 10 independent services on the same feed URL, we will
   hit the URL 10 times. This is similar to how we currently do 1 webhook/service, so it's plausible that in
   the future we will want to have some kind of generic caching layer.
 - The polling interval is divorced from the actual feed repoll time. Ideally we would schedule the goroutine
   only when we need it, rather than checking frequently, determining we have nothing to do, and going back
   to sleep.
 - The `Clients` instance is hackily added via a `SetClients` global. Given we *know* which client the service always wants, I wonder if we should be injecting this into the Service somehow.

### DIFF
--- a/src/github.com/matrix-org/go-neb/services/feedreader/feedreader.go
+++ b/src/github.com/matrix-org/go-neb/services/feedreader/feedreader.go
@@ -2,12 +2,14 @@ package services
 
 import (
 	"errors"
+	"fmt"
 	log "github.com/Sirupsen/logrus"
 	"github.com/matrix-org/go-neb/database"
 	"github.com/matrix-org/go-neb/matrix"
 	"github.com/matrix-org/go-neb/polling"
 	"github.com/matrix-org/go-neb/types"
 	"github.com/mmcdole/gofeed"
+	"html"
 	"time"
 )
 
@@ -16,7 +18,7 @@ const minPollingIntervalSeconds = (10 * 60) // 10min
 type feedPoller struct{}
 
 func (p *feedPoller) IntervalSecs() int64 { return 10 }
-func (p *feedPoller) OnPoll(s types.Service) {
+func (p *feedPoller) OnPoll(s types.Service, cli *matrix.Client) {
 	logger := log.WithFields(log.Fields{
 		"service_id":   s.ServiceID(),
 		"service_type": s.ServiceType(),
@@ -40,13 +42,13 @@ func (p *feedPoller) OnPoll(s types.Service) {
 
 	// Query each feed and send new items to subscribed rooms
 	for _, u := range pollFeeds {
-		items, err := p.queryFeed(frService, u)
+		feed, items, err := p.queryFeed(frService, u)
 		if err != nil {
 			logger.WithField("feed_url", u).WithError(err).Error("Failed to query feed")
 			continue
 		}
 		for _, i := range items {
-			if err := p.sendToRooms(frService, u, i); err != nil {
+			if err := p.sendToRooms(frService, cli, u, feed, i); err != nil {
 				logger.WithFields(log.Fields{
 					"feed_url":   u,
 					log.ErrorKey: err,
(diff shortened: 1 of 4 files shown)


src/github.com/matrix-org/go-neb/services/feedreader/feedreader.go
```diff
@@ -0,0 +1,235 @@
...
+func (p *feedPoller) IntervalSecs() int64 { return 10 }
+func (p *feedPoller) OnPoll(s types.Service, cli *matrix.Client) {
+	logger := log.WithFields(log.Fields{
+		"service_id":   s.ServiceID(),
+		"service_type": s.ServiceType(),
+	})
+
+	frService, ok := s.(*feedReaderService)
+	if !ok {
+		logger.Error("FeedReader: OnPoll called without an Feed Service instance")
```
lukebarnard1: s/an/a

src/github.com/matrix-org/go-neb/services/feedreader/feedreader.go
```diff
@@ -0,0 +1,235 @@
...
+		return
+	}
+	now := time.Now().Unix() // Second resolution
+
+	// Work out which feeds should be polled
+	var pollFeeds []string
+	for u, feedInfo := range frService.Feeds {
```
lukebarnard1: It took me awhile to figure out that ```u``` is a URL, despite that probably being a Go convention(?)

I also didn't realise ```pollFeeds``` was an array of URLs for awhile.

src/github.com/matrix-org/go-neb/services/feedreader/feedreader.go
```diff
@@ -0,0 +1,235 @@
...
+					"feed_url":   u,
+					log.ErrorKey: err,
+					"item":       item,
+				}).Error("Failed to send item to room")
+			}
+		}
+	}
+
+	// Persist the service to save the next poll times if we did some queries
+	if len(pollFeeds) == 0 {
```
lukebarnard1: I feel like this early return should go before the range over ```pollFields```

src/github.com/matrix-org/go-neb/services/feedreader/feedreader.go
```diff
@@ -0,0 +1,235 @@
...
+			logger.WithError(err).WithField("room_id", roomID).Error("Failed to send to room")
+		}
+	}
+	return nil
+}
+
+// SomeOne posted a new article: Title Of The Entry ( https://someurl.com/blag )
+func itemToHTML(feed *gofeed.Feed, item gofeed.Item) matrix.HTMLMessage {
+	return matrix.GetHTMLMessage("m.notice", fmt.Sprintf(
+		"<i>%s</i> posted a new article: %s ( %s )",
```
lukebarnard1: Will it always be "someone posting a new article"? If anything, we can only assume that a feed has been updated. I feel like this should be configurable somewhere...

src/github.com/matrix-org/go-neb/services/feedreader/feedreader.go
```diff
@@ -0,0 +1,235 @@
...
+func (s *feedReaderService) ServiceID() string     { return s.id }
+func (s *feedReaderService) ServiceType() string   { return "feedreader" }
+func (s *feedReaderService) Poller() types.Poller  { return &feedPoller{} }
+
+// Register will check the liveness of each RSS feed given. If all feeds check out okay, no error is returned.
+func (s *feedReaderService) Register(oldService types.Service, client *matrix.Client) error {
+	if len(s.Feeds) == 0 {
+		// this is an error UNLESS the old service had some feeds in which case they are deleting us :(
+		var numOldFeeds int
+		oldFeedService, ok := oldService.(*feedReaderService)
```
lukebarnard1: Not sure what this syntax is...

src/github.com/matrix-org/go-neb/services/feedreader/feedreader.go
```diff
@@ -0,0 +1,235 @@
...
+	return nil
+}
+
+func (s *feedReaderService) PostRegister(oldService types.Service) {
+	if len(s.Feeds) == 0 { // bye-bye :(
+		logger := log.WithFields(log.Fields{
+			"service_id":   s.ServiceID(),
+			"service_type": s.ServiceType(),
+		})
+		logger.Info("Deleting service (0 feeds)")
```
lukebarnard1: ```0 feeds``` might look like 0 is a formatted value. What about, "no feeds remaining"?

src/github.com/matrix-org/go-neb/services/feedreader/feedreader.go
```diff
@@ -0,0 +1,235 @@
...
+		polling.StopPolling(s)
+		if err := database.GetServiceDB().DeleteService(s.ServiceID()); err != nil {
+			logger.WithError(err).Error("Failed to delete service")
+		}
+	}
+}
+
+func init() {
+	types.RegisterService(func(serviceID, serviceUserID, webhookEndpointURL string) types.Service {
```
lukebarnard1: Not sure what ```webhookEndpointURL```  is

src/github.com/matrix-org/go-neb/services/feedreader/feedreader.go
Kegsay: It's a Go convention to be terse with variable names (they prefer single letters or just a few).

src/github.com/matrix-org/go-neb/services/feedreader/feedreader.go
Kegsay: Agreed.

src/github.com/matrix-org/go-neb/services/feedreader/feedreader.go
Kegsay: We don't currently support templates in Go-NEB. I think we will want to in the future, so people can configure their GH messages, RSS messages, JIRA messages, etc. But that doesn't exist yet.

src/github.com/matrix-org/go-neb/services/feedreader/feedreader.go
Kegsay: Not relevant to this particular service, but that is the function signature of the factory method.

src/github.com/matrix-org/go-neb/services/feedreader/feedreader.go
Kegsay: Done.

src/github.com/matrix-org/go-neb/services/feedreader/feedreader.go
Kegsay: Done.

src/github.com/matrix-org/go-neb/services/feedreader/feedreader.go
lukebarnard1: Ah, ok.

Kegsay:
Anything left to do here?

Kegsay:
IRL'd, nope!

PR #80, opened 2016-10-10
Rejig how feedreaders stores rooms to send updates to

Kegsay:
- Log fatal errors when `ListenAndServe`ing (e.g. port in use).
- Move `Rooms` to inside the `Feed` object.
- Check that there is at least 1 room to send updates to.


NegativeMjark:
LGTM

PR #81, opened 2016-10-10
s/feedreader/rssbot/g

Kegsay:
Because apparently "Feed Reader" is the name of a thing popular enough that we
don't want to step on toes, so let's call it the name of a less popular thing.


NegativeMjark:
LGTM

PR #82, opened 2016-10-11
Move OnPoll into the Service method set

Kegsay:
This feels a lot better because now `OnPoll` works in a similar way to
`OnReceiveWebhook` (called on a `Service`) rather than have this strange
global-per-service-type struct.

This also now prevents waking up a goroutine every 10s as the `OnPoll` interface now returns *when* it wants to next be polled.


NegativeMjark:
LGTM

PR #83, opened 2016-10-11
Add a caching HTTP client

Kegsay:
Most of the RSS feeds I `curl`'d had HTTP cache headers (sometimes to say, don't cache at all, other times to put a low `max-age`), so a generic RFC-compliant HTTP caching library will be Good Enough for our use cases. E.g:

```
> GET /news/rss.xml?edition=uk HTTP/1.1
> Host: feeds.bbci.co.uk
> User-Agent: curl/7.43.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Server: nginx
< Content-Type: text/xml; charset=utf-8
< X-Content-Type-Options: nosniff
< x-origin-route: xrt-ext
< Content-Length: 28853
< X-Cache-Action: HIT
< X-Cache-Hits: 2
< X-Cache-Age: 8
< Cache-Control: max-age=51
< Date: Tue, 11 Oct 2016 14:38:24 GMT
< Connection: keep-alive
<
```

I verified that the caching library is doing its job via `tcpdump`, and indeed it seems to be!

### DIFF
--- a/src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
+++ b/src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	log "github.com/Sirupsen/logrus"
+	"github.com/die-net/lrucache"
 	"github.com/gregjones/httpcache"
 	"github.com/matrix-org/go-neb/database"
 	"github.com/matrix-org/go-neb/matrix"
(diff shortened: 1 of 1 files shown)


src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
```diff
@@ -214,6 +220,7 @@ func itemToHTML(feed *gofeed.Feed, item gofeed.Item) matrix.HTMLMessage {
 }
 
 func init() {
+	cachingClient = httpcache.NewMemoryCacheTransport().Client()
```
NegativeMjark: Are there any limits on how much memory this cache will consume?

NegativeMjark:
LGTM

src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
Kegsay: No, though it looks like we can plug-and-play an [LruCache](https://github.com/die-net/lrucache).

Kegsay: Done.

NegativeMjark:
LGTM

PR #84, opened 2016-10-13
Maintain a list of GUIDs per feed and use them to de-dupe

Kegsay:
Some RSS feeds will edit the published time of an item AFTER putting it out,
which resulted in RSS Bot sending the same article twice. We now remember the
"GUID" field for each item and de-dupe based on that. Normal timestamp
algorithm still applies.


NegativeMjark:
LGTM

Issue #63, opened 2016-09-08
Locked database errors


Kegsay:
```
{
  "error":"database is locked",
  "level":"error",
  "msg":"Failed to persist next_batch token",
  "next_batch":"s39513903_25872915_34252_3708100_14886_10006_12",
  "time":"2016-09-08T09:06:24+01:00",
  "user_id":"@neb_giphy:matrix.org"
}
```

Kegsay:
https://github.com/mattn/go-sqlite3/issues/274

PR #85, opened 2016-10-13
Edit README to put `_busy_timeout` to fix database is locked errors (#63)

Kegsay:
It's crap we have to do this, but I guess in the future we'll use Postgres? :/

Related:
- https://github.com/matrix-org/go-neb/issues/63
- https://github.com/mattn/go-sqlite3/issues/274#issuecomment-232942571


NegativeMjark:
LGTM

PR #87, opened 2016-10-14
Instrument all HTTP handlers

Kegsay:
It compiles, and curling `/metrics` returns stuff which looks plausible soooo yay?


NegativeMjark:
LGTM

PR #91, opened 2016-10-17
Add metrics to NEB

Kegsay:
In addition to the raw HTTP-level metrics, we now increment counters for:
 - Incoming webhook events which are recognised as belonging to a service (add `service_type` as label).
 - Incoming `/configureService` requests which were succesful (add `service_type` as label).
 - Incoming `!command` requests which succeeded/failed (add `status` and `cmd` as labels).
 - Outgoing RSS polls (add `http_status` and `url` as labels).

Also fixed Guggy's bad error handling on non-2xx responses.

### DIFF
--- a/src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
+++ b/src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
@@ -14,6 +14,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"html"
 	"net/http"
+	"net/url"
 	"time"
 )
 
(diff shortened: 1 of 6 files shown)


src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
```diff
@@ -150,6 +161,25 @@ func (s *rssBotService) OnPoll(cli *matrix.Client) time.Time {
 	return s.nextTimestamp()
 }
 
+func sendMetric(urlStr string, err error) {
```
NegativeMjark: maybe call this ``incrementMetrics`` or something?

NegativeMjark:
other than odd function names LGTM.

src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
Kegsay: Done.

PR #92, opened 2016-10-18
Set "Go-NEB" as the User-Agent on RSS feed polling

### LINKED ISSUE #88: RSS: Reddit rate limits on the default Golang UA
Reddit heavily rate limits their RSS feeds for common non-browser UA strings (e.g. curls, Golangs, Pythons). They don't apply such a strict rate limit for custom UAs/browser UAs. It's so heavy, it means that basically every request 429s (even if you wait 1 minute, 2 minutes, etc). We should set a UA string to benefit from the higher rate limit (which is 1 per ~5 sec).

Kegsay:
Some services, notably Reddit, basically force you to set a custom UA in order
to use their RSS feeds. All the common default non-browser UAs are HEAVILY
rate-limited such that you can't realistically use them.

Fixes #88

### DIFF
--- a/src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
+++ b/src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
@@ -306,9 +306,20 @@ func itemToHTML(feed *gofeed.Feed, item gofeed.Item) matrix.HTMLMessage {
 	))
 }
 
+type userAgentRoundTripper struct {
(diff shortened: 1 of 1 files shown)


NegativeMjark:
LGTM. Might be nice to set this for all the HTTP requests GO-NEB makes. Might be nice to include a version somewhere.

Kegsay:
Filed as https://github.com/matrix-org/go-neb/issues/93

PR #100, opened 2016-10-24
Prevent panicking code from taking down the entire process

### LINKED ISSUE #99: Isolate services
If one service code panics, it takes down the entire process. We should defer/recover.

Kegsay:
Panicking poll loops and incoming Matrix events could cause the entire process to fall over. They are now protected by a `defer`/`recover` block.

Also, whilst all incoming HTTP requests are, by default, already protected from `panics`, they were falling back to the standard output logger rather than being caught by logrus, so we now catch them in a similar way in order to:
 - Log to logrus.
 - Return a 500 rather than abruptly closing the connection.
 - Return a JSON error response.

Manually tested by introducing NPEs at various points in the code. Fixes #99.

### DIFF
--- a/src/github.com/matrix-org/go-neb/polling/polling.go
+++ b/src/github.com/matrix-org/go-neb/polling/polling.go
@@ -5,6 +5,7 @@ import (
 	"github.com/matrix-org/go-neb/clients"
 	"github.com/matrix-org/go-neb/database"
 	"github.com/matrix-org/go-neb/types"
+	"runtime/debug"
 	"sync"
 	"time"
 )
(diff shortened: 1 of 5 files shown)


NegativeMjark:
LGTM

NegativeMjark:
Although it'd awesome if you had some tests for the Protect function.

src/github.com/matrix-org/go-neb/polling/polling.go
```diff
@@ -71,6 +72,15 @@ func pollLoop(service types.Service, ts int64) {
...
 		"service_type": service.ServiceType(),
 	})
+
+	defer func() {
+		if r := recover(); r != nil {
+			logger.WithField("panic", r).Errorf(
+				"pollLoop panicked!\n%s", debug.Stack(),
+			)
+		}
+	}()
```
NegativeMjark: Might want to add metrics or something to track when things die.

Killing the loop might be the right thing here since it's likely that whatever cause us to panic will keep panic'ing

Maybe add some comments to explain what the intended behaviour is?

Kegsay: Thought about it, and no, it doesn't. I'm treating panicking code as unrecoverable, and not executing the exact same code again hoping for a different outcome.

Kegsay: Didn't get your 2nd comment when I first replied. Will add comments.

NegativeMjark:
LGTM

PR #101, opened 2016-10-25
Make it possible to do integration tests

Kegsay:
This means:
 - Allow an alternative HTTP client to be used for all Matrix clients so we can mock out the Homeserver.
 - Allow an alternative `ServeMux` to be used so we can inject fake Go-NEB requests to things like `/configureService`, and check their responses.
 - Allow alternative environment variables to be specified rather than exclusively by `os.Getenv`, so we can use an in-memory database.

Added a noddy `TestConfigureClient` test to make sure that:
 - We can send `/configureClient` requests.
 - We can validate the response.
 - We can assert `/sync` streams are started.

The design of the test **will change**, this is simply a PoC.


src/github.com/matrix-org/go-neb/goneb_services_test.go
```diff
@@ -0,0 +1,103 @@
...
+		"Sync":true,
+		"AutoJoinRooms":true
+	}`))
+	mux.ServeHTTP(mockWriter, mockReq)
+	expectCode := 200
+	if mockWriter.Code != expectCode {
+		t.Errorf("TestConfigureClient wanted HTTP status %d, got %d", expectCode, mockWriter.Code)
+	}
+
+	<-syncChan
```
NegativeMjark: What happens if syncChan isn't written to?

Kegsay: It will block forever. Any wait to do timeouts?

src/github.com/matrix-org/go-neb/goneb_services_test.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func TestConfigureClient(t *testing.T) {
+	for k := range tripper.handlers {
+		delete(tripper.handlers, k)
+	}
+	mockWriter := httptest.NewRecorder()
+	tripper.Handle("POST", "/_matrix/client/r0/user/@link:hyrule/filter",
+		func(req *http.Request) (*http.Response, error) {
+			return newResponse(200, `{
```
NegativeMjark: Do we need to check if this request happens?

Kegsay: We need to return the filter ID or `/sync` will never happen.

src/github.com/matrix-org/go-neb/goneb_services_test.go
NegativeMjark: ``go test`` takes a timeout flag.

```
 -timeout t
            If a test runs longer than t, panic.
            The default is 10 minutes (10m).
```

Kegsay: 10 *minutes*?! Okay, will tweak the pre-commit hook to do a sane time like 5s.

Kegsay: https://github.com/constabulary/gb/blob/065d2b2b1ba151d68c65f718c911bd4947f2cc13/cmd/gb/test.go#L47 ....

Kegsay: Oh wait, `gb test -timeout 5s` works

NegativeMjark: Alternatively you could roll your own with https://golang.org/pkg/time/#AfterFunc

Kegsay: Done using `-timeout 5s`

src/github.com/matrix-org/go-neb/goneb_services_test.go
```diff
@@ -0,0 +1,103 @@
...
+	handlers map[string]func(req *http.Request) (*http.Response, error)
+}
+
+func (rt MockTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	key := req.Method + " " + req.URL.Path
+	h := rt.handlers[key]
+	if h == nil {
+		panic(
+			"Test RoundTrip: Unhandled request: " + key + "\n" +
+				"Handlers: " + strconv.Itoa(len(rt.handlers)),
```
Kegsay: Done

src/github.com/matrix-org/go-neb/goneb_services_test.go
NegativeMjark: Something like
```go
timer := time.AfterFunc(10 * time.Second), func() { t.Fatal("Timeout") })
// test some things
timer.stop()
```

Kegsay:
@NegativeMjark PTAL

NegativeMjark:
LGTM

PR #102, opened 2016-10-26
Move test util functions to dedicated file to clean up test files

Kegsay:
Also use verbose mode on tests so you can see which test is running when things
print to stdout.


NegativeMjark:
LGTM

PR #104, opened 2016-10-27
Make ServiceDB an interface called `Storer`

Kegsay:
Implement a nop `Storer` called `NopStorage` which tests can make use of.
This was awful to write.

### DIFF
--- a/src/github.com/matrix-org/go-neb/database/db.go
+++ b/src/github.com/matrix-org/go-neb/database/db.go
@@ -15,8 +15,6 @@ type ServiceDB struct {
 }
 
 // A single global instance of the service DB.
-// XXX: I can't think of any way of doing this without one without creating
-//      cyclical dependencies somewhere -- Kegan
 var globalServiceDB Storer
 
 // SetServiceDB sets the global service DB instance.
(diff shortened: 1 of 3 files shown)


src/github.com/matrix-org/go-neb/database/interface.go
```diff
@@ -0,0 +1,140 @@
...
+
+	LoadBotOptions(userID, roomID string) (opts types.BotOptions, err error)
+	StoreBotOptions(opts types.BotOptions) (oldOpts types.BotOptions, err error)
+
+	InsertFromConfig(cfg *api.ConfigFile) error
+}
+
+// NopStorage nops every store API call. This is intended to be embedded into derived structs
+// in tests
+type NopStorage struct{}
```
NegativeMjark: Oh, I guess you want to be able to import it from other places. Still might be nice to put it somewhere else rather than having it right next to the interface definition? YMMV

src/github.com/matrix-org/go-neb/database/db.go
```diff
@@ -17,15 +17,15 @@ type ServiceDB struct {
 // A single global instance of the service DB.
 // XXX: I can't think of any way of doing this without one without creating
 //      cyclical dependencies somewhere -- Kegan
```
NegativeMjark: (I realise this isn't actually part of the diff. But could you elaborate on what you wanted to do and why it resulted in cycles?)

NegativeMjark:
LGTM

src/github.com/matrix-org/go-neb/database/interface.go
Kegsay: As per IRL, having them there is as good a place as any.

PR #106, opened 2016-11-01
Split up the enormous handlers.go into more manageable files

Kegsay:
Split them up based on the HTTP API they are implementing. This makes it a lot easier to see which piece of code will execute when random HTTP endpoints are hit. This also means I can remove the `Handler` suffix on all the structs since it is part of the `handlers` package now.


NegativeMjark:
LGTM

PR #109, opened 2016-11-02
Delete API info from README and point to gh-page. Add gendoc script

Kegsay:


gendoc.sh
```diff
@@ -0,0 +1,26 @@
...
+
+DOC_DIR=godoc
+
+# Run a godoc server which we will scrape. Clobber the GOPATH to include
+# only our dependencies.
+GOPATH=$(pwd):$(pwd)/vendor godoc -http=localhost:6060 &
+DOC_PID=$!
+
+# Wait for the server to init
+sleep 1
```
NegativeMjark: Might be nice to put a while loop here in case the godoc server takes longer to start?

gendoc.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
NegativeMjark: "set -eu" please

gendoc.sh
NegativeMjark: You could use a simple wget to check if the server is listening yet.

NegativeMjark:
Otherwise LGTM

gendoc.sh
Kegsay: Done

gendoc.sh
Kegsay: I've `set -u` since `wget` returns `8` ( https://www.gnu.org/software/wget/manual/html_node/Exit-Status.html ) which is fine, but I can't then check `$?` because the script terminates before I can. In addition, I'm relying on `curls` exit code being non-zero whilst testing if the server is up.

PR #111, opened 2016-11-03
Remove Plugin. Replace with Commands() and Expansions()

### LINKED ISSUE #19: Service.Plugin(roomID) feels weird
The steps for invoking plugin code involve:
- Receive an event
- Load `Services` in the given room.
- f.e `Service` call `Plugin(roomID)`
- f.e `Plugin` call the right thing (command/expand/whatever)

I don't understand why we don't just have `Service.Plugin()`. All the callbacks already include the `roomID` so we don't even use the fact that we're in a closure to pull out the room ID. In fact, even the whole concept of `Plugin` I just _don't get_, why don't we just have:
- `Service.Commands() []Command`
- `Service.Expansions() []Expansion`

The actual implementations always rely on the fact that they are in a closure around the `Service` - they can't work independently, so I don't see the benefit of having the indirection of a `Plugin` concept.

Kegsay:
Also split out `types.go` into separate files by the area they cover (services, auth, actions) so they can be more easily found.

Fixes #19


NegativeMjark:
LGTM

PR #112, opened 2016-11-03
Add default implementations for ServiceType(), ServiceID() and ServiceUserID()

Kegsay:
In addition to making there be less boilerplate, this also has the nice side-effect of shutting up golint warnings \o/

Will convert the rest of the services once this format has been OKed.


NegativeMjark:
LGTM

PR #114, opened 2016-11-04
Remove docs from the README and link to the godoc instead

Kegsay:
Also add example configs for each service.


NegativeMjark:
LGTM

PR #115, opened 2016-11-04
400 if a service which needs a syncing client is configured without one

Kegsay:
This catches a common configuration error and fails early. Also, remove `roomID` from `Commands()` and `Expansions()` as we never used it.

### DIFF
--- a/src/github.com/matrix-org/go-neb/api/api.go
+++ b/src/github.com/matrix-org/go-neb/api/api.go
@@ -61,8 +61,8 @@ type ClientConfig struct {
 	HomeserverURL string
 	// The matrix access token to authenticate the requests with.
 	AccessToken string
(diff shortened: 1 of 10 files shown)


NegativeMjark:
Maybe s/syncing client/client with "Sync" enabled/ ?
LGTM.

Kegsay:
I chose the wording carefully because I don't want the API for `/configureClient` to be leaking through to error messages in `/configureService`. I would like to `lower_case_and_underscores` all the APIs eventually, which would make the error message here misleading. The concept of a "syncing" client is a safer bet, and is unlikely to change however, hence why I chose that specific wording. WDYT?

NegativeMjark:
Would it make sense for api.ClientConfig.Sync to use the phrase "syncing client" in its documentation?

Kegsay:
Yes, good idea. Will do that instead.

PR #116, opened 2016-11-04
Implement slackapi Service

Kegsay:
WIP


src/github.com/matrix-org/go-neb/services/slackapi/message.go
```diff
@@ -84,6 +85,7 @@ var netClient = &http.Client{
 	Timeout: time.Second * 10,
 }
 
+// TODO: What does this do?
 var linkRegex, _ = regexp.Compile("<([^|]+)(\\|([^>]+))?>")
```
aviraldg: Helps us find and convert [Slack links](https://api.slack.com/docs/message-formatting#linking_to_urls)

almereyda:
How could this solve Prometheus and GitLab integration, is this even mimicking the Slack API on Matrix nodes?
Or would one need to connect the desired services to Slack and only then to Matrix?

config.sample.yaml
```diff
@@ -108,3 +108,12 @@ services:
               Events: ["push", "issues"]
             "matrix-org/dendron":
               Events: ["pull_request"]
+
+  - ID: "slackapi_service"
+    Type: "slackapi"
+    UserID: "@slackapi:localhost" # requires a Syncing client
```
Kegsay: Does this require a syncing client? Does it ever respond to matrix users?

src/github.com/matrix-org/go-neb/services/slackapi/slackapi.go
```diff
@@ -38,11 +38,14 @@ func (s *Service) OnReceiveWebhook(w http.ResponseWriter, req *http.Request, cli
 
 	slackMessage, err := getSlackMessage(*req)
 	if err != nil {
+		log.WithFields(log.Fields{"slackMessage":slackMessage, "err":err}).Print("Slack message error")
```
Kegsay: Please log errors with `.Error` not `.Print`.

src/github.com/matrix-org/go-neb/services/slackapi/slackapi.go
```diff
@@ -38,11 +38,14 @@ func (s *Service) OnReceiveWebhook(w http.ResponseWriter, req *http.Request, cli
...
 		w.WriteHeader(500)
 		return
 	}
+
 	htmlMessage, err := slackMessageToHTMLMessage(slackMessage)
 	if err != nil {
+		log.WithField("err", err).Error("Converting slack message to HTML")
```
Kegsay: Also known as:
```go
log.WithError(err).Error("Converting slack message to HTML")
```

You should never specify the error key directly, use `log.ErrorKey` instead.

src/github.com/matrix-org/go-neb/services/slackapi/slackapi.go
Kegsay: Also, you should never specify the error key directly, use `log.ErrorKey` instead.

src/github.com/matrix-org/go-neb/services/slackapi/slackapi.go
```diff
@@ -53,6 +56,20 @@ func (s *Service) OnReceiveWebhook(w http.ResponseWriter, req *http.Request, cli
...
+	for _, mapping := range s.Hooks {
+		if _, err := client.JoinRoom(mapping.RoomID, "", ""); err != nil {
+			log.WithFields(log.Fields{
+				log.ErrorKey: err,
+				"room_id":    mapping.RoomID,
+				"user_id":    client.UserID,
+			}).Error("Failed to join room")
+		}
+	}
+	return nil
```
Kegsay: You need to tell the caller what the webhook endpoint is. See https://github.com/matrix-org/go-neb/blob/master/src/github.com/matrix-org/go-neb/services/travisci/travisci.go#L245

lukebarnard1: But there is no 1-1 mapping between slackapi service instances and hooks; one service could have many hooks... Given `webhookEndpointURL := baseURL + "services/hooks/" + base64ServiceID`, I think it would make more sense to not have any concept of hook IDs and instead just have a single room ID stored per service. The service ID can then be used as a replacement for this hook ID concept.

Kegsay: Not following. The person who makes a `slackapi` needs to be given a URL they can tell their Slack apps to use. In Go-NEB, you have to use the webhook URL given to your service, no other URL will map to your service otherwise.

lukebarnard1: As I understand it, for webhooks in Go-NEB, the last path component is the base64 encoded service ID. The path must begin with `/services/hooks` and I have tested that `/services/hooks/$HOOK_ID/$SERVICE_ID` is handled correctly. 

So you _can_ have multiple webhooks directed at a single service, and the `webhookEndpointURL` provided _is_ an example of a hook that will hit the service, but it does not include a `HookID`. 

So there are two options: 1. remove the hook ID as a concept, and just have one webhook per slackapi service or 2. just send the existing `webhookEndpointURL` without the hook ID and expect the client to fill it in.

Kegsay: Let's not re-invent the architecture. Please do #1 which is exactly what Travis has been doing for months with success.

config.sample.yaml
```diff
@@ -111,7 +111,6 @@ services:
 
   - ID: "slackapi_service"
     Type: "slackapi"
-    UserID: "@slackapi:localhost" # requires a Syncing client
```
Kegsay: I think you're confused. You still need a `UserID`, else who will send the messages as they arrive from the webhook? It just doesn't need to hit `/sync` to receive `!commands`. You just needed to remove the comment.

lukebarnard1: I thought the personal bots are the ones sending messages into rooms?

src/github.com/matrix-org/go-neb/services/slackapi/slackapi.go
lukebarnard1: ok, fair enough

config.sample.yaml
Kegsay: Go-NEB has no concept of personal or global bots. It's just a user who sends messages. You need to give a user ID.

lukebarnard1: Oh yes, sorry getting projects all muddled

src/github.com/matrix-org/go-neb/services/slackapi/slackapi.go
```diff
@@ -14,11 +14,11 @@ const ServiceType = "slackapi"
 
 type Service struct {
 	types.DefaultService
-	// maps from hookID -> roomID
-	Hooks map[string]struct {
-		RoomID      string `json:"room_id"`
-		MessageType string `json:"message_type"`
-	}  `json:"hooks"`
+	webhookEndpointURL string
+	// The URL which should be added to .travis.yml - Populated by Go-NEB after Service registration.
```
Kegsay: Not travis 0:)

Kegsay:
Otherwise, LGTM!

Kegsay:
Otherwise, LGTM!

src/github.com/matrix-org/go-neb/services/slackapi/slackapi.go
```diff
@@ -5,23 +5,37 @@ import (
...
 
+// Service contains the Config fields for the Slack API service.
+//
+// This service will send HTML formatted messages into a room when an outgoing slack webhook
+// hits WebhookURL.
+//
+// Example JSON request:
+//   {
+//       "room_id": "!someroomid:some.domain.com",
+//		 "message_type": "m.text"
```
Kegsay: Indentation.

Kegsay:
LGTM

PR #123, opened 2016-11-18
Fix #59 : Make pling commands case-insensitive

### LINKED ISSUE #59: pling commands should probably be case insensitive

Kegsay:


NegativeMjark:
LGTM

PR #129, opened 2016-11-24
Add a dedicated testutils package

Kegsay:
In the spirit of "if you have to do something 3 times then factor it
out", make a testutils package to put all the `RoundTrip` boilerplate.

I don't overly like having test packages, especially mixed in with
code, but I don't see a nicer way of doing this without ending up
with a sprawling mess of copypasta'd test boilerplate which will
be an absolute nightmare to maintain. This is only going to get
worse as I want to add more tests with exactly this API shape.

I think this is the lesser of two evils.

### DIFF
--- a/src/github.com/matrix-org/go-neb/services/travisci/travisci_test.go
+++ b/src/github.com/matrix-org/go-neb/services/travisci/travisci_test.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/matrix-org/go-neb/database"
 	"github.com/matrix-org/go-neb/matrix"
+	"github.com/matrix-org/go-neb/testutils"
 	"github.com/matrix-org/go-neb/types"
 )
 
@@ -91,14 +92,6 @@ var travisTests = []struct {
 	},
 }
 
-type MockTransport struct {
-	roundTrip func(*http.Request) (*http.Response, error)
-}
-
-func (t MockTransport) RoundTrip(req *http.Request) (*http.Response, error) {
-	return t.roundTrip(req)
-}
-
 func TestTravisCI(t *testing.T) {
 	database.SetServiceDB(&database.NopStorage{})
 
@@ -106,8 +99,8 @@ func TestTravisCI(t *testing.T) {
 	urlToKey := make(map[string]string)
 	urlToKey["https://api.travis-ci.org/config"] = travisOrgPEMPublicKey
 	urlToKey["https://api.travis-ci.com/config"] = travisComPEMPublicKey
-	travisTransport := struct{ MockTransport }{}
-	travisTransport.roundTrip = func(req *http.Request) (*http.Response, error) {
+	travisTransport := struct{ testutils.MockTransport }{}
+	travisTransport.RT = func(req *http.Request) (*http.Response, error) {
 		if key := urlToKey[req.URL.String()]; key != "" {
 			escKey, _ := json.Marshal(key)
 			return &http.Response{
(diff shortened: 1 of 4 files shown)


NegativeMjark:
Hmm. I'm not entirely sure if this buys anything other than a sense of cleanliness.

I disagree that test boilerplate is necessarily a nightmare to maintain. If anything tests are easier to maintain when all the code is self contained and changes can be made to one test without affecting all of them.

src/github.com/matrix-org/go-neb/testutils/testutils.go
```diff
@@ -0,0 +1,15 @@
+package testutils
+
+import (
+	"net/http"
+)
+
+// MockTransport implements RoundTripper
+type MockTransport struct {
+	RT func(*http.Request) (*http.Response, error)
```
NegativeMjark: If you are going to split this out to a library then you should probably add some comments to explain what's going on.

Kegsay:
My reasoning is that there are a lot of "common" actions I want to do like intercept HTTP requests and send back HTTP responses. 

> changes can be made to one test without affecting all of them

These are so "low-level" that I disagree that they will be changed per-test, I meant there's nothing really *to* change.

Kegsay:
As an example for why I want a single place for boilerplate, consider integration testing of services.

The test will spin up a brand new Go-NEB and configure the service and then send some events from Matrix to the service. Every single one of these tests needs to handle the boilerplate of setting up a filter (200 OK with the right JSON object shape), and the /sync loop. Your proposal from what I can tell is to duplicate this for every test package. If we then decide to change the sync code to not use a filter we then need to update every test package rather than just the util package.

WDYT?

NegativeMjark:
> Furthermore, the boilerplate undeniably obfuscates the intent of the tests: it adds a lot of noise.

Unfortunately unless you are careful shifting the code to a library will obfuscate what the code does.
I'm not convinced that you are going to be able to adequately express your intent in the code. 

> Every single one of these tests needs to handle the boilerplate of setting up a filter (200 OK with the right JSON object shape), and the /sync loop.

If you have shared boilerplate like this then I'm fairly happy for it to be in a testutils package, since presumably it's trying to get past the same bit of shared code in GO-NEB, and therefore will need to be changed in bulk if that bit of shared code changes.

Kegsay:
So you don't mind sharing code wrt filter/sync loop but do mind wrt http req/res handling? The justification for the latter being that it's clearer if it's just in the test?

src/github.com/matrix-org/go-neb/services/travisci/travisci_test.go
```diff
@@ -124,8 +117,8 @@ func TestTravisCI(t *testing.T) {
 
 	// Intercept message sending to Matrix and mock responses
 	msgs := []matrix.TextMessage{}
-	matrixTrans := struct{ MockTransport }{}
-	matrixTrans.roundTrip = func(req *http.Request) (*http.Response, error) {
+	matrixTrans := struct{ testutils.MockTransport }{}
+	matrixTrans.RT = func(req *http.Request) (*http.Response, error) {
```
Kegsay: Done

src/github.com/matrix-org/go-neb/testutils/testutils.go
Kegsay: Done

Kegsay:
From OOB chat:

> I think as long as the utility functions are sufficiently clear as to what they do it's fine.

@NegativeMjark PTAL

PR #130, opened 2016-11-24
Fix #72: Support smart quotes

### LINKED ISSUE #72: Parse smart-quotes for commands

Kegsay:


NegativeMjark:
LGTM. Although I think that "smart" needs to be in quote marks.

PR #132, opened 2016-11-25
Use Giphy's translate API for better accuracy

Kegsay:


NegativeMjark:
LGTM although it might be better to decode the result to `interface{}` to check if it's the empty list.

Kegsay:
Given how rubbish Giphy's API is, I doubt that it will always return `{ data: [] }` if there are 0 results (e.g. if it fails to query because it is overloaded). Rather than failing with obscure JSON decode errors, I feel it is safer to assume that there are 0 results, hence the catch all.

PR #135, opened 2016-11-29
Redo how recent GUIDs are calculated

### LINKED ISSUE #133: RSS: Can send old news
Reported by seequ. For example, http://www.bittium.com/docs/templates/stat/rss_news.php?lang=en (It keeps the entire history.) and https://feeds.finance.yahoo.com/rss/2.0/headline?s=aapl&region=US&lang=en-US

Kegsay:
This is an attempt to fix #133.

Previously, we just clobbered the recent GUIDs with the lastest response
every single time, assuming that Atom/RSS feeds would consistently return the
same items. This appears to not be the case. In the wild, the number of items
returned on a single request can vary (sometimes even being 1 or 2 when usually
it is 50!).

This patch alters *how many* and *which* GUIDs we keep between requests, in an
attempt to prevent sending old news for buggy RSS feeds.

### DIFF
--- a/src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
+++ b/src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
@@ -297,6 +297,7 @@ func (s *Service) queryFeed(feedURL string) (*gofeed.Feed, []gofeed.Item, error)
 	lastSet := uniqueStrings(f.RecentGUIDs) // e.g. [4,5,6]
 	thisSet := uniqueGuids(feed.Items)      // e.g. [1,2,3]
 	guids := append(thisSet, lastSet...)    // e.g. [1,2,3,4,5,6]
+	guids = uniqueStrings(guids)
 	if len(guids) > maxGuids {
 		// Critically this favours the NEWEST elements, which are the ones we're most likely to see again.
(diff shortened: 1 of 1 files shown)


src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
```diff
@@ -289,26 +279,33 @@ func (s *Service) queryFeed(feedURL string) (*gofeed.Feed, []gofeed.Item, error)
...
+	if len(guids) > maxGuids {
+		// Critically this favours the NEWEST elements, which are the ones we're most likely to see again.
+		guids = guids[0:maxGuids]
 	}
 
 	// Update the service config to persist the new times
 	f.NextPollTimestampSecs = nextPollTsSec
 	f.FeedUpdatedTimestampSecs = now
-	f.RecentGUIDs = guids
+	f.RecentGUIDs = uniqueStrings(guids)
```
NegativeMjark: Should this last ``uniq`` be before the ``len(guids) > maxGuids`` check?

Kegsay: Yes. Done.

NegativeMjark:
LGTM

PR #136, opened 2016-11-29
Don't use gofeed.ParseURL

Kegsay:
It leaks the response because it doesn't close the `resp.Body` on non-2xx.


NegativeMjark:
LGTM

NegativeMjark:
Maybe comment on why we aren't using ParseURL?

Kegsay:
Done.

PR #137, opened 2016-12-01
Use gomatrix

Kegsay:
Horrible diff, I'm sorry.


src/github.com/matrix-org/go-neb/clients/clients.go
```diff
@@ -343,35 +319,49 @@ func (c *Clients) onRoomMemberEvent(client *matrix.Client, event *matrix.Event)
...
+		syncer.OnEventType("m.room.member", func(event *gomatrix.Event) {
 			c.onRoomMemberEvent(client, event)
 		})
 	}
 
 	if config.Sync {
-		go client.Sync()
+		go func() {
+			for {
+				if e := client.Sync(); e != nil {
```
NegativeMjark: Does it ever stop syncing?

Kegsay: No, well spotted. When `StopSync()` is called there is a `nil` error which should result in it stopping then.

Kegsay: Done.

PR #158, opened 2017-02-20
Imgur image search bot

rxl881:


Kegsay:
So. Many. Crosses. Please fix Travis :smile:

src/github.com/matrix-org/go-neb/services/imgur/imgur.go
```diff
@@ -0,0 +1,270 @@
...
+	MP4         string `json:"mp4"`         // mp4	string	OPTIONAL, The direct link to the .mp4. Only available if the image is animated and type is 'image/gif'.
+	MP4Size     int64  `json:"mp4_size"`    // mp4_size	integer	OPTIONAL, The Content-Length of the .mp4. Only available if the image is animated and type is 'image/gif'. Note that a zero value (0) is possible if the video has not yet been generated
+	Looping     bool   `json:"looping"`     // looping	boolean	OPTIONAL, Whether the image has a looping animation. Only available if the image is animated and type is 'image/gif'.
+	NSFW        bool   `json:"nsfw"`        // nsfw	boolean	Indicates if the image has been marked as nsfw or not. Defaults to null if information is not available.
+	Topic       string `json:"topic"`       // topic	string	Topic of the gallery image.
+	Section     string `json:"section"`     // section	string	If the image has been categorized by our backend then this will contain the section the image belongs in. (funny, cats, adviceanimals, wtf, etc)
+	IsAlbum     bool   `json:"is_album"`    // is_album	boolean	If it's an album or not
+	// ** Uninplemented fields **
```
Kegsay: > Uninplemented

src/github.com/matrix-org/go-neb/services/imgur/imgur.go
```diff
@@ -0,0 +1,270 @@
...
+// ServiceType of the Imgur service
+const ServiceType = "imgur"
+
+var httpClient = &http.Client{}
+
+type imgurGalleryImage struct {
+	ID          string `json:"id"`          // id	string	The ID for the image
+	Title       string `json:"title"`       // title	string	The title of the image.
+	Description string `json:"description"` // description	string	Description of the image.
+	DateTime    int64  `json:"datetime"`    // datetime	integer	Time inserted into the gallery, epoch time
```
Kegsay: The comments which restate the name of the key and the type is redundant.

src/github.com/matrix-org/go-neb/services/imgur/imgur.go
```diff
@@ -0,0 +1,270 @@
...
+func (s *Service) cmdImgurImgSearch(client *gomatrix.Client, roomID, userID string, args []string) (interface{}, error) {
+
+	if len(args) < 1 {
+		return usageMessage(), nil
+	}
+
+	// Get the query text to search for.
+	querySentence := strings.Join(args, " ")
+
+	searchResultImage, searchResultAlbum, err := s.text2imgImgur(querySentence)
```
Kegsay: `text2imgImgur` : Redundant names are redundant. I know it's to do with imgur, this is the name of the package. Ditto for this function `cmdImgurImgSearch`.

src/github.com/matrix-org/go-neb/services/imgur/imgur.go
```diff
@@ -0,0 +1,270 @@
...
+	if searchResultImage != nil {
+		var imgURL = searchResultImage.Link
+		if imgURL == "" {
+			return gomatrix.TextMessage{
+				MsgType: "m.notice",
+				Body:    "No image found!",
+			}, nil
+		}
+
+		// FIXME -- Sometimes upload fails with a cryptic error - "msg=Upload request failed code=400"
```
Kegsay: Redundant FIXME now?

src/github.com/matrix-org/go-neb/services/imgur/imgur.go
```diff
@@ -0,0 +1,270 @@
...
+// text2imgImgur returns info about an image or an album
+func (s *Service) text2imgImgur(query string) (*imgurGalleryImage, *imgurGalleryAlbum, error) {
+	log.Info("Searching Imgur for an image of a ", query)
+
+	query = url.QueryEscape(query)
+	var base = "https://api.imgur.com/3/gallery/search"
+	var sort = "time"  // time | viral | top
+	var window = "all" // day | week | month | year | all
+	var page = 1
+	var urlString = fmt.Sprintf("%s/%s/%s/%d?q=%s", base, sort, window, page, query)
```
Kegsay: I like the `sort`, `window` and `page` vars which are used to add more information about how the URL is constructed. I don't like the `base`, which adds nothing. Can you please inline `base`?

src/github.com/matrix-org/go-neb/services/imgur/imgur.go
```diff
@@ -0,0 +1,270 @@
...
+
+	req.Header.Add("Authorization", "Client-ID "+s.ClientID)
+	res, err := httpClient.Do(req)
+	if res != nil {
+		defer res.Body.Close()
+	}
+	if err != nil {
+		return nil, nil, err
+	}
+	if res.StatusCode > 200 {
```
Kegsay: Technically it should be 2xx is OK.

rxl881: :)

src/github.com/matrix-org/go-neb/services/imgur/imgur.go
rxl881: Fair enough - I had left them in to match the comments for the unimplemented fields (where it probably makes sense to retain the field name and type, for when we implement them in future).

rxl881:
Hmm... I'm not sure what is causing the travis build to fail. Build and all tests are fine in dev.

Details of the failed travis builds show it complaining about punctuation and spaces in various files (that I haven't modified). Has something more general failed @Kegsay ?
screen shot 2017-02-20 at 18 29 34

Kegsay:
as per RL: it's the cyclomatic complexity tool on `imgur.go:191`.

src/github.com/matrix-org/go-neb/services/imgur/imgur.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+type imgurGalleryAlbum struct {
+	ID          string              `json:"id"`           // The ID for the album
+	Title       string              `json:"title"`        // The title of the album.
+	Description string              `json:"description"`  // Description of the album.
+	DateTime    int64               `json:"datetime"`     // Time inserted into the gallery, epoch time
+	Views       int64               `json:"views"`        // The number of album views
+	Link        string              `json:"link"`         // The URL link to the album
+	NSFW        bool                `json:"nsfw"`         // Indicates if the album has been marked as nsfw or not. Defaults to null if information is not available.
```
Kegsay: Not that we actually do anything with this, but for tri-state flags like this (true/false/null) you'll need to use `*bool` otherwise `null` will be treated as `false`. https://play.golang.org/p/US6V2qiykX

src/github.com/matrix-org/go-neb/services/imgur/imgur.go
```diff
@@ -0,0 +1,268 @@
...
+	// points	integer	Upvotes minus downvotes
+	// score	integer	Imgur popularity score
+	// vote	string	The current user's vote on the album. null if not signed in or if the user hasn't voted on it.
+	// favorite	boolean	Indicates if the current user favorited the album. Defaults to false if not signed in.
+	// comment_count	int	Number of comments on the gallery album.
+	// topic_id	integer	Topic ID of the gallery album.
+}
+
+type imgurSearchResponse struct {
+	Data    []json.RawMessage `json:"data"`
```
Kegsay: Explain why you're doing this and not giving it a concrete struct type.

src/github.com/matrix-org/go-neb/services/imgur/imgur.go
```diff
@@ -0,0 +1,268 @@
...
+	if len(args) < 1 {
+		return usageMessage(), nil
+	}
+
+	// Get the query text to search for.
+	querySentence := strings.Join(args, " ")
+
+	searchResultImage, searchResultAlbum, err := s.text2img(querySentence)
+
+	if err != nil {
```
Kegsay: Style: Please reduce the number of new lines. Structure your code into paragraphs as if it were prose, doing one task in one "paragraph". E.g:
```go
// sanity check
if len(args) < 1 {
    return usageMessage(), nil
}

// extract and submit query
querySentence := strings.Join(args, " ")
searchResultImage, searchResultAlbum, err := s.text2img(querySentence)
if err != nil {
    return nil, err
}

// process response
// ...
```

There will always be a degree of subjectivity here, but there shouldn't be cases where you start a function with a blank line (like on `:138`). Whilst `go fmt` doesn't know/care about this, I find it annoying since it damages readability as I don't know where a single logical action starts or stops without having to read the entire function and having well-defined breaks is useful in a similar vein to why people like having short sentences for clarity rather than a really long run on sentence like this one where you don't really know where my point is going to end and when I'm going to start a new point.

src/github.com/matrix-org/go-neb/services/imgur/imgur.go
```diff
@@ -0,0 +1,268 @@
...
+// text2img returns info about an image or an album
+func (s *Service) text2img(query string) (*imgurGalleryImage, *imgurGalleryAlbum, error) {
+	log.Info("Searching Imgur for an image of a ", query)
+
+	query = url.QueryEscape(query)
+
+	var sort = "time"  // time | viral | top
+	var window = "all" // day | week | month | year | all
+	var page = 1
+	var urlString = fmt.Sprintf("%s/%s/%s/%d?q=%s", "https://api.imgur.com/3/gallery/search", sort, window, page, query)
```
Kegsay: When I asked you to in-line, I was more going for:
```go
fmt.Sprintf("https://api.imgur.com/3/gallery/search/%s/%s/%d?q=%s", sort, window, page, query)
```

:)

src/github.com/matrix-org/go-neb/services/imgur/imgur.go
```diff
@@ -0,0 +1,268 @@
...
+	Success bool              `json:"success"` // Request completed successfully
+	Status  int               `json:"status"`  // HTTP response code
+}
+
+// Service contains the Config fields for the Imgur service.
+//
+// Example request:
+//   {
+//			"client_id": "AIzaSyA4FD39..."
+//			"client_secret": "ASdsaijwdfASD..."
```
Kegsay: Nit: Probably just use 1 tab :)

src/github.com/matrix-org/go-neb/services/imgur/imgur_test.go
```diff
@@ -0,0 +1,121 @@
...
+		}, nil
+	})
+	// clobber the imgur service http client instance
+	httpClient = &http.Client{Transport: imgurTrans}
+
+	// Create the imgur service
+	srv, err := types.CreateService("id", ServiceType, "@imgurbot:hyrule", []byte(
+		`{
+			"client_id":"`+clientID+`"
+		}`,
```
Kegsay: Use:
```go
fmt.Sprintf(`{
    "client_id":"%s"
}`, clientID)
```

instead.

src/github.com/matrix-org/go-neb/services/imgur/imgur.go
rxl881: Nice tip! Thanks :) ... I'll modify it for if / when we use it in future.

src/github.com/matrix-org/go-neb/services/imgur/imgur.go
```diff
@@ -248,6 +223,45 @@ func (s *Service) text2img(query string) (*imgurGalleryImage, *imgurGalleryAlbum
 	return nil, nil, fmt.Errorf("No images found")
 }
 
+// Query imgur and return HTTP response or error
+func queryImgur(query string, s *Service) ([]byte, error) {
```
Kegsay: Just pass the client ID rather than the service.

src/github.com/matrix-org/go-neb/services/imgur/imgur.go
rxl881: I've added a comment to explain that the response data is a mix of imgurGalleryImage and imgurGalleryAlbum objects (to be parsed at a later date). Hopefully this is what you were thinking of?

src/github.com/matrix-org/go-neb/services/imgur/imgur.go
rxl881: Fair enough. I've cleaned up and added some more comments.

rxl881:
PTAL

src/github.com/matrix-org/go-neb/services/imgur/imgur.go
```diff
@@ -208,7 +208,7 @@ func (s *Service) text2img(query string) (*imgurGalleryImage, *imgurGalleryAlbum
 	var images []imgurGalleryImage
 	for i := 0; i < len(searchResults.Data); i++ {
 		var image imgurGalleryImage
-		if err := json.Unmarshal(searchResults.Data[i], &image); err == nil && !image.IsAlbum {
+		if err := json.Unmarshal(searchResults.Data[i], &image); err == nil && !*(image.IsAlbum) {
```
Kegsay: Careful! You can't dereference and check it like this. What happens if it was `nil`? https://play.golang.org/p/oDN14Z9vqd
Instead, do:
```go
if image.IsAlbum != nil && !*(image.IsAlbum) {
    // ...
}
```

src/github.com/matrix-org/go-neb/services/imgur/imgur.go
Kegsay: Yup!

Kegsay:
```
imgur_test.go:65: cannot use true (type bool) as type *bool in field value
```

src/github.com/matrix-org/go-neb/services/imgur/imgur.go
```diff
@@ -0,0 +1,288 @@
...
+	ClientSecret string `json:"client_secret"`
+}
+
+// Commands supported:
+//    !imgur some_search_query_without_quotes
+// Responds with a suitable image into the same room as the command.
+func (s *Service) Commands(client *gomatrix.Client) []types.Command {
+	return []types.Command{
+		types.Command{
+			Path: []string{"imgur", "help"},
```
Kegsay: An interesting point here is that this makes it impossible to search on Imgur for the word "help" :) Just a random thought, nothing for you to do.

Kegsay:
Other than failing Travis and nil dereference, LGTM.

src/github.com/matrix-org/go-neb/services/imgur/imgur.go
rxl881: Yeah, I did realise that ... and then decided it wasn't a problem. I think that you could probably search for "help!" if that's what you're really after (seems to work).

rxl881:
Just fixing the travis integration issue ... what you get for committing and running out of the office (sorry) :/

PR #162, opened 2017-02-23
Use util.JSONResponse

Kegsay:
Horrible diff, I'm sorry.


src/github.com/matrix-org/go-neb/services/jira/webhook/webhook.go
```diff
@@ -153,22 +155,18 @@ func createWebhook(jrealm *jira.Realm, webhookEndpointURL, userID string) error
 	return err
 }
 
-func getWebhook(cli *gojira.Client, webhookEndpointURL string) (*jiraWebhook, *util.HTTPError) {
+func getWebhook(cli *gojira.Client, webhookEndpointURL string) (*jiraWebhook, bool, error) {
```
NegativeMjark: I think you should add some comments to explain that the bool returned means "forbidden".

NegativeMjark:
Otherwise than commenting on the bool return LGTM.

(The diff is actually pretty reasonable to review because  `nil, &util.HTTPError{nil, ` and `util.MessageResponse(400, ` have the same number of characters.)

PR #188, opened 2017-07-02
Dockerfile and docker image build script

superdump:
It would be nice if we could make some official matrix-org docker image out of this that we update with every commit and release with appropriate version tags.


Kegsay:
Assuming it works, LGTM

Issue #195, opened 2017-08-14
Add moar GitHub commands.


t3chguy:
+ +1
+ -1
+ comment
+ assign
+ close

uhoreg:
comment command is https://github.com/matrix-org/go-neb/issues/149

Matrixcoffee:
!search

t3chguy:
!tag

MTRNord:
!label (?)

t3chguy:
s/tag/label/

t3chguy:
heloo there foobar - testing from t3ch-neb

t3chguy:
heloo there foobar - testing from t3ch-neb numero DOS

t3chguy:
- [ ] !github label
- [ ] !github search

ara4n:
(the quickest way for go-neb to progress atm is going to be from the community, so: patches welcome!! for all this. agreed it'd kick ass :)

t3chguy:
@ara4n `!github comment` was already merged, some more here: https://github.com/matrix-org/go-neb/pull/197

t3chguy:
Testing !github comment after some refactoring

uhoreg:
search command is https://github.com/matrix-org/go-neb/issues/55

Matrixcoffee:
!reopen

t3chguy:
@Matrixcoffee as soon as you close https://github.com/matrix-org/go-neb/issues/149 ;)

Matrixcoffee:
:D

Yes master.

t3chguy:
@Matrixcoffee here you go: https://github.com/matrix-org/go-neb/pull/206

PR #197, opened 2017-08-22
Moar GitHub commands

t3chguy:
[For #195] adds
+ !github react
+ !github close
+ !github assign

refactors usage strings bc DRY

### DIFF
--- a/src/github.com/matrix-org/go-neb/services/github/github.go
+++ b/src/github.com/matrix-org/go-neb/services/github/github.go
@@ -77,14 +77,15 @@ func (s *Service) requireGithubClientFor(userID string) (cli *gogithub.Client, r
 	return
 }
 
+const cmdGithubCreateUsage = `!github create owner/repo "issue title" "description"`
+
 func (s *Service) cmdGithubCreate(roomID, userID string, args []string) (interface{}, error) {
 	cli, resp, err := s.requireGithubClientFor(userID)
 	if cli == nil {
 		return resp, err
 	}
 	if len(args) == 0 {
-		return &gomatrix.TextMessage{"m.notice",
-			`Usage: !github create owner/repo "issue title" "description"`}, nil
+		return &gomatrix.TextMessage{"m.notice", "Usage: " + cmdGithubCreateUsage}, nil
 	}
 
 	// We expect the args to look like:
@@ -97,14 +98,12 @@ func (s *Service) cmdGithubCreate(roomID, userID string, args []string) (interfa
 		// look for a default repo
 		defaultRepo := s.defaultRepo(roomID)
 		if defaultRepo == "" {
-			return &gomatrix.TextMessage{"m.notice",
-				`Usage: !github create owner/repo "issue title" "description"`}, nil
+			return &gomatrix.TextMessage{"m.notice", "Usage: " + cmdGithubCreateUsage}, nil
 		}
 		// default repo should pass the regexp
 		ownerRepoGroups = ownerRepoRegex.FindStringSubmatch(defaultRepo)
 		if len(ownerRepoGroups) == 0 {
-			return &gomatrix.TextMessage{"m.notice",
-				`Malformed default repo. Usage: !github create owner/repo "issue title" "description"`}, nil
+			return &gomatrix.TextMessage{"m.notice", "Malformed default repo. Usage: " + cmdGithubCreateUsage}, nil
 		}
 
 		// insert the default as the first arg to reuse the same indices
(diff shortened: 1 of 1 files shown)


src/github.com/matrix-org/go-neb/services/github/github.go
```diff
@@ -202,21 +200,127 @@ func (s *Service) cmdGithubComment(roomID, userID string, args []string) (interf
...
+}
+
+func (s *Service) getIssueDetailsFor(input, roomID, usage string) (owner, repo string, issueNum int, resp interface{}) {
+	// We expect the input to look like:
+	// "[owner/repo]#issue"
+	// They can omit the owner/repo if there is a default one set.
+	// Look for a default if the first arg is just an issue number
+	ownerRepoIssueGroups := ownerRepoIssueRegexAnchored.FindStringSubmatch(input)
+
+	if len(ownerRepoIssueGroups) != 5 {
```
ara4n: shouldn't we check the string itself rather than just its length?

src/github.com/matrix-org/go-neb/services/github/github.go
```diff
@@ -142,55 +141,54 @@ func (s *Service) cmdGithubCreate(roomID, userID string, args []string) (interfa
 	return gomatrix.TextMessage{"m.notice", fmt.Sprintf("Created issue: %s", *issue.HTMLURL)}, nil
 }
 
-func (s *Service) cmdGithubComment(roomID, userID string, args []string) (interface{}, error) {
+const cmdGithubReactUsage = `!github react [owner/repo]#issue (+1|-1|laugh|confused|heart|hooray)`
```
t3chguy: as in someone posts an issue and we just do `:+1:`
that'd require rewriting most of the service to be aware of the most recent issue created

I'm just using the strings that the API lists as valid

src/github.com/matrix-org/go-neb/services/github/github.go
t3chguy: ownerRepoIssueGroups is a group of regex results

src/github.com/matrix-org/go-neb/services/github/github.go
t3chguy: (Plus it is what it refers to them in the UI)
image

ara4n: i wasn't suggesting a literal `:+1:` to upvote - more that if the example syntax gave the emoji in terms of their emoji-one colon form, then Riot users could either type:

`!github react matrix-org/synapse#1234 :+1:`
or
`!github react matrix-org/synapse#1234 +1`
or
`!github react matrix-org/synapse#1234 👍`
and they would all work.

Perhaps we can include uncertain & grin (or their : codes?) for the other ambiguous ones?  Or at least the : codes?

t3chguy: so accept the official terms and add aliases, sure

t3chguy: Added their formal names, their formal names wrapped in `:`'s and their emoji variants as aliases

PTAL @ara4n

src/github.com/matrix-org/go-neb/services/github/github.go
```diff
@@ -77,14 +78,69 @@ func (s *Service) requireGithubClientFor(userID string) (cli *gogithub.Client, r
...
+
+	summarizedIssues := searchResult.Issues[0:numberOfSummaries]
+
+	htmlBuffer.WriteString(fmt.Sprintf("Found %d results, here are the most relevant:<br><ol>", numResults))
+	plainBuffer.WriteString(fmt.Sprintf("Found %d results, here are the most relevant:\n", numResults))
+	for i, issue := range summarizedIssues {
+		if issue.HTMLURL == nil || issue.User.Login == nil || issue.Title == nil {
+			continue
+		}
+		htmlBuffer.WriteString(fmt.Sprintf(`<li><a href="%s" rel="noopener">%s: %s</a></li>`, *issue.HTMLURL, *issue.User.Login, *issue.Title))
```
Kegsay: This is vulnerable to XSS attacks if you put arbitrary HTML as the issue title. Please escape appropriately.

src/github.com/matrix-org/go-neb/services/github/github.go
```diff
@@ -77,14 +78,69 @@ func (s *Service) requireGithubClientFor(userID string) (cli *gogithub.Client, r
...
+		return nil, fmt.Errorf("Failed to search. HTTP %d", res.StatusCode)
+	}
+
+	if searchResult.Total == nil || *searchResult.Total == 0 {
+		return &gomatrix.TextMessage{"m.notice", "No results found for your search query!"}, nil
+	}
+
+	numResults := *searchResult.Total
+	var htmlBuffer bytes.Buffer
+	var plainBuffer bytes.Buffer
```
Kegsay: So far away from the first usage on `:116` :(

src/github.com/matrix-org/go-neb/services/github/github.go
t3chguy: done

src/github.com/matrix-org/go-neb/services/github/github.go
t3chguy: done

src/github.com/matrix-org/go-neb/services/github/github.go
```diff
@@ -77,14 +79,69 @@ func (s *Service) requireGithubClientFor(userID string) (cli *gogithub.Client, r
...
+
+	summarizedIssues := searchResult.Issues[0:numberOfSummaries]
+
+	var htmlBuffer bytes.Buffer
+	var plainBuffer bytes.Buffer
+	htmlBuffer.WriteString(fmt.Sprintf("Found %d results, here are the most relevant:<br><ol>", numResults))
+	plainBuffer.WriteString(fmt.Sprintf("Found %d results, here are the most relevant:\n", numResults))
+	for i, issue := range summarizedIssues {
+		if issue.HTMLURL == nil || issue.User.Login == nil || issue.Title == nil {
+			continue
```
Kegsay: It's unfortunate that we apply the max issue check and then have this additional check which may result in even fewer entries. I'd be tempted to just loop all the `searchResult.Issues` and stop once you've added `numberGithubSearchSummaries` or reached the end of the slice.

Kegsay:
Otherwise LGTM

src/github.com/matrix-org/go-neb/services/github/github.go
t3chguy: done

PR #213, opened 2017-12-12
Implement CircleCI Webhook Service

### LINKED ISSUE #192: Add CircleCi integration
A lot of people use circleci.com instead of travis, and it would be helpful for those people to have a circleci integration!

They should have a webhook api like Travis so it shouldn't be too hard to add.

MTRNord:
This PR will fix #192 

## TODO
- [x] Test it on a actual Project
- [x] Add all other possible template options provided in the payload
- [x] Rename the template fields to match their names in the payload (Leave the TravisCI template names too?) (Auto generated map for payload does this. TravisCI namings are still supported with Deprecation warning inside the documentation)
- [x] Write Tests
- [x] Add a Underscore to the map keys of the payload instead doing them all lowercase

Signed-Off-By: Marcel Radzio <mtrnord1@gmail.com>

---

After this is merged I would appreciate if before or after the next go-neb update on the official server it would get listed on Scalar ;) Otherwise non power user will not notice/use this. (Of course I cant do that as it is closed source, But I propably will do a dimension PR for it later.)


MTRNord:
I am not sure how this happens https://travis-ci.org/matrix-org/go-neb/builds/315403483#L470 as this basicly is the same test as the TravisCI one.

MTRNord:
grafik

MTRNord:
You may wonder about the new lib structs: This is used to auto Map the Struct. It might be possible to add it as a function directly if wanted. Also as it is of type "map[string]interface{}" I do convert it for strings and int types to "map[string]string". time.Time typed values are currently missing and will be added.

MTRNord:
Bump :) Anyone who wants to review this? (I tested it and it works fine. Like the Travis one basicly)

src/github.com/matrix-org/go-neb/services/circleci/circleci.go
```diff
@@ -0,0 +1,270 @@
...
+
+// DefaultTemplate contains the template that will be used if none is supplied.
+// This matches the default mentioned at: https://docs.travis-ci.com/user/notifications#Customizing-slack-notifications
+const DefaultTemplate = (`%{repository_slug}#%{build_num} (%{branch} - %{commit} : %{committer_name}): %{outcome}
+	Build details : %{build_url}`)
+
+// Matches 'owner/repo'
+var ownerRepoRegex = regexp.MustCompile(`^([A-z0-9-_.]+)/([A-z0-9-_.]+)$`)
+
+var httpClient = &http.Client{}
```
t3chguy: unused variable

src/github.com/matrix-org/go-neb/services/circleci/types.go
```diff
@@ -0,0 +1,44 @@
+package circleci
+
+import "time"
+
+// WebhookNotification is the response of a webhook notification by circleCI
+type WebhookNotification struct {
```
t3chguy: should be unexported as travisci's is

src/github.com/matrix-org/go-neb/services/circleci/circleci.go
```diff
@@ -0,0 +1,270 @@
...
+
+	if !p.StartTime.IsZero()  && !p.StopTime.IsZero() {
+			t["duration"] = p.StopTime.Sub(p.StartTime).String()
+			t["elapsed_time"] = t["duration"]
+	}
+
+	t["build_url"] = p.BuildURL
+
+	// Make the full struct data into the Map without manual mapping
+	rawT := structs.Map(p)
```
t3chguy: I'm not a fan of this *magic*
best to hardcode all the conversions if anything then for consistency

t3chguy: (plus its what the travisci service does)

src/github.com/matrix-org/go-neb/services/circleci/circleci.go
```diff
@@ -0,0 +1,270 @@
...
+			"Body":  req.Body,
+			log.ErrorKey: err,
+		}).Warn("Failed to Read Body")
+		w.WriteHeader(403)
+		return
+	}
+	//fmt.Printf("%s\n", body)
+
+	var notif WebhookNotification
+	if err := json.Unmarshal(body, &notif); err != nil {
```
t3chguy: You can replace this and the `ioutil` call with a `err = json.NewDecoder(req.Body).Decode(&notify)` as per https://stackoverflow.com/a/21198571/3532235

src/github.com/matrix-org/go-neb/services/circleci/util.go
```diff
@@ -0,0 +1,61 @@
...
+	b.r = append(b.r, b.runeBytes[0:n]...)
+}
+
+func (b *buffer) indent() {
+	if len(b.r) > 0 {
+		b.r = append(b.r, '_')
+	}
+}
+
+func CamelCaseToUnderscore(s string) string {
```
t3chguy: this is part of the magic I'm not a fan of (obviously this bit is less magical but still not a fan)

t3chguy: actually this whole file is for that magic

MTRNord:
Thanks for the Review @t3chguy i will work on them Next weekend :)

src/github.com/matrix-org/go-neb/services/circleci/circleci.go
MTRNord: Hm any Idea how I manualy should map the Steps Array?

MTRNord:
ping @t3chguy any chance reviewing the new commit in the next week or so? :)

t3chguy:
I shall give this an indepth review and test rn

t3chguy:
Simply for namespacing it'd be great if it kept the trend as github has with `webhook-circleci` and circleci_webhook.go etc so that webhook services are clearly namespaced, I'll review the rest of the PR as a whole as this will just be a trivial rename/replace

src/github.com/matrix-org/go-neb/services/circleci/circleci.go
```diff
@@ -0,0 +1,276 @@
...
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/matrix-org/go-neb/database"
+	"github.com/matrix-org/go-neb/types"
+	"github.com/matrix-org/gomatrix"
+	"strconv"
+)
+
+// ServiceType of the CircleCI service.
+const ServiceType = "circleci"
```
t3chguy: webhook-circleci

src/github.com/matrix-org/go-neb/services/circleci/circleci.go
```diff
@@ -0,0 +1,276 @@
...
+//           "!ewfug483gsfe:localhost": {
+//               repos: {
+//                   "matrix-org/go-neb": {
+//                       template: "%{repository_slug}#%{build_num} (%{branch} - %{commit} : %{committer_name}): %{outcome}\nBuild details : %{build_url}"
+//                   }
+//               }
+//           }
+//       }
+//   }
+type Service struct {
```
t3chguy: WebhookService (why not :P)

src/github.com/matrix-org/go-neb/services/circleci/circleci.go
```diff
@@ -0,0 +1,276 @@
...
+	t["lifecycle"] = p.Lifecycle
+	t["outcome"] = p.Outcome
+	t["status"] = p.Status
+	switch value := p.RetryOf.(type) {
+	case string:
+		t["retry_of"] = value
+	case int:
+		t["retry_of"] = strconv.Itoa(value)
+	}
+	//TODO: Figure out how to map the Steps Slice/Array
```
t3chguy: Care to elaborate?

src/github.com/matrix-org/go-neb/services/circleci/circleci.go
MTRNord: github has the webhook as suffix. I will keep the Github syntax

t3chguy: and `WebhookServiceType` for consistency :P

MTRNord: Also travisCI isn't namespaced as a webhook

vendor/manifest
```diff
@@ -76,6 +76,12 @@
 			"revision": "0e55cff0d6054f900b91c0b5355fe6a6966fad4b",
 			"branch": "master"
 		},
+		{
```
t3chguy: this needs removing from manifest

MTRNord:
Yes the CircleCI docs are awful. Mostly because v2 is pretty new (and horrible broken sometimes). And yes it doesn't have a way to verify them either.

t3chguy:
Surely you could verify them by requesting the build details in a request?

MTRNord:
@t3chguy I guess I can do that but wouldn't slow down the whole service a lot?

t3chguy:
Github and Travis both verify so I'd argue an extra second or two of delay in a webhook is more accepted than an exploitable one. Until https://discuss.circleci.com/t/add-webhook-secret-token-like-github/5185/7 anyway

MTRNord:
@t3chguy I just checked the CircleCI Docs: It seems any request to the API needs a API token that is written on the Users settings page. So this would need to get this API Token from the User first to verify via the API

t3chguy:
argh that makes it less than ideal :/

MTRNord:
@t3chguy I can only check what VCS and which User they use on Github or Bitbucket for the Project without that API Token but I fully understand what you mean.

src/github.com/matrix-org/go-neb/services/circleci/circleci.go
MTRNord: Ping @t3chguy ^

t3chguy: yeah ignore this because of github-webhooks being a specialcase

Cadair:
@MTRNord maybe I am misunderstanding, but at least for public repos you appear to be able to hit the circle CI API to get build info unauthenticated, I am looking up the artifcats [here](https://github.com/Cadair/giles/blob/master/giles/blueprints/circleci.py#L48)

(arguably for private repos the security thing is less important because you don't know what the webhook endpoint is?)

src/github.com/matrix-org/go-neb/services/circleci/circleci.go
```diff
@@ -0,0 +1,276 @@
...
+// If the repository matches a known Git* repository, a notification will be formed from the
+// template for that repository and a notice will be sent to Matrix.
+//
+// Go-NEB cannot register with CircleCI for webhooks automatically. The user must manually add the
+// webhook endpoint URL to their .circleci/config.yml file:
+//    notify:
+//		webhooks:
+//			- url: https://example.com/services/hooks/circle
+//
+// See https://circleci.com/docs/1.0/configuration/#notify for more information.
```
turt2live: note that the circlci 2.0 docs completely omit the webhooks documentation, but based on user reports it seems to still work. Not really sure what they're doing :/

MTRNord:
@turt2live yep I dont think they will ever add something to verify webhooks :/ I dont care of this gets closed or merged. Tbh I dont even use circle Co anymore as their Page bringst my cpu to 100% (but that's another Story)

PR #227, opened 2018-04-18
Avoid double logging

michaelkaye:
Currently, if LOG_DIR is set, we log everything to stderr and additionally log to the three files in LOG_DIR.

This should disable logging to stderr if we are explicitly setting a LOG_DIR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ BIND_ADDRESS=:4050 DATABASE_TYPE=sqlite3 DATABASE_URL=go-neb.db?_busy_timeout=50
  - `DATABASE_URL` is where to find the database file. One will be created if it does not exist. It is a URL so parameters can be passed to it. We recommend setting `_busy_timeout=5000` to prevent sqlite3 "database is locked" errors.
  - `BASE_URL` should be the public-facing endpoint that sites like Github can send webhooks to.
  - `CONFIG_FILE` is the path to the configuration file to read from. This isn't included in the example above, so Go-NEB will operate in HTTP mode.
-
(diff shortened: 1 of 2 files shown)


michaelkaye:
https://github.com/Sirupsen/logrus/issues/328 made this as the suggestion for removing logging when you have a hook doing most of the work.

README.md
```diff
@@ -105,7 +105,7 @@ BIND_ADDRESS=:4050 DATABASE_TYPE=sqlite3 DATABASE_URL=go-neb.db?_busy_timeout=50
  - `DATABASE_URL` is where to find the database file. One will be created if it does not exist. It is a URL so parameters can be passed to it. We recommend setting `_busy_timeout=5000` to prevent sqlite3 "database is locked" errors.
  - `BASE_URL` should be the public-facing endpoint that sites like Github can send webhooks to.
  - `CONFIG_FILE` is the path to the configuration file to read from. This isn't included in the example above, so Go-NEB will operate in HTTP mode.
-
+ - `LOG_DIR` is a directory that log files will be written to, with log rotation enabled. If set, logging to stderr will be disabled.
```
jcgruenhage: Surely logrus would default to logging to stdout, not stderr, right? So setting this should disable both logging to stdout and stderr.

michaelkaye: From https://github.com/sirupsen/logrus/blob/master/README.md#example
```
  // Output to stdout instead of the default stderr
  // Can be any io.Writer, see below for File example
  log.SetOutput(os.Stdout)
```

jcgruenhage: Weird.. but doesn't matter, since we log to "Discard" now.
Good to know though, I'll keep it in mind.

PR #233, opened 2018-06-06
Update Docker instructions to be more automated-build friendly

turt2live:
This PR also assumes that the image will end up on docker.io (as an automated build).


Dockerfile
```diff
@@ -1,8 +1,21 @@
+# Build go-neb
+FROM golang:1.8-alpine as builder
```
jcgruenhage: why 1.8-alpine and not 1.10-alpine?

.dockerignore
```diff
@@ -0,0 +1,10 @@
+vendor/pkg
+vendor/src
+pkg
+hooks
+bin
+
+# Other stuff we just don't need
+.gitignore
+.git
+.travis.yml
```
jcgruenhage: iirc the .dockerignore file is only important for not having to hand a lot of data to the docker daemon over http. I agree that .git and some of the directories up there should definitely be in there, but ignoring .gitignore and .travis.yml seems a little random

README.md
```diff
@@ -225,7 +225,12 @@ sensible-browser http://localhost:6060/pkg/github.com/matrix-org/go-neb
 
 ## Docker image
 
-There is a `Dockerfile` in the root of the repository and a `build-docker-image.sh` script that uses an alpine-based golang container to build `go-neb` (note that this will overwrite host-built binaries in `pkg/` and `bin/`) and then builds the docker image using that binary.
+To get started quickly, use the image from docker.io:
+```
+docker run -v /path/to/data:/data -e "BASE_URL=http://your.public.url:4050" matrixdotorg/go-neb
+```
+
+If you'd prefer to build the file yourself, clone this repository and build the `Dockerfile`.
```
jcgruenhage: A lot better than before, thanks!

.dockerignore
turt2live: They were picked randomly, so I'll remove them

Dockerfile
turt2live: Copy/pasta from the old script - will update

PR #243, opened 2018-08-08
Remove "url" from RSS metric

Half-Shot:
It's not really that useful.


michaelkaye:
What does deployment look like for this change in metric names? Do we need to change any prometheus alerts, or do some summary of historical data?

(+/- any build failures ;)

erikjohnston:
Nothing in internal config seems to reference the time series, so I don't think there is anything to do. Prometheus should be happy with suddenly dropping a label, I believe.

PR #249, opened 2018-10-01
alertmanager support

jcgruenhage:


anoadragon453:
I just meant could it be a security risk ( not sure what data alert manager can produce and if it's sensitive).

jcgruenhage:
I don't think it's sending any sensitive data, but that toally depends on how people configure it

anoadragon453:
@jcgruenhage is it worth it to add a config option to disable it?

jcgruenhage:
I don't think so. The bot has access to the messages it sends anyway, since it has a token of a user that is in that room. Also, I can't think of a threatmodel where you'd trust go-neb with the data but not it's logs.. Specifically, for people using it through scalar it will end up on matrix.org anyway, and for people who host it themselves they can just set the loglevel to get rid of it.

Half-Shot:
I hope this was tested, btw?

jcgruenhage:
Yes, it was

MTRNord:
Just as I dont understand what this does :) What does this in case of Go-Neb? Is it Matrix->Prometheus or the other way around or did I missunderstood it completly?

turt2live:
It's for alertmanager -> matrix

MTRNord:
@turt2live ok thanks!

PR #253, opened 2018-11-26
Update build system

michaelkaye:
Update to go 1.11, for both travis tests and docker test/builds.

Run linter as part of docker build (but doesn't end up in the end result container)

Remove linting errors so we start from a clean build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,18 @@
 # Build go-neb
-FROM golang:1.10-alpine as builder
+FROM golang:1.11-alpine as builder
+
+RUN apk add --no-cache -t build-deps git gcc musl-dev go
 
 COPY . /tmp/go-neb
 WORKDIR /tmp/go-neb
-RUN apk add --no-cache -t build-deps git gcc musl-dev go \
-    && go get -u github.com/constabulary/gb/... \
+RUN go get -u github.com/constabulary/gb/... \
+    && go get github.com/golang/lint/golint \
+    && go get github.com/fzipp/gocyclo \
     && gb vendor restore \
     && gb build -f github.com/matrix-org/go-neb
 
+# Ensures we're lint-free
+RUN /tmp/go-neb/hooks/pre-commit
 
 # Run go-neb
 FROM alpine:3.7
--- a/src/github.com/matrix-org/go-neb/services/alertmanager/alertmanager.go
+++ b/src/github.com/matrix-org/go-neb/services/alertmanager/alertmanager.go
@@ -52,7 +52,7 @@ type Service struct {
 	} `json:"rooms"`
 }
 
-// The payload from Alertmanager
+// WebhookNotification is the payload from Alertmanager
 type WebhookNotification struct {
 	Version           string            `json:"version"`
 	GroupKey          string            `json:"groupKey"`
@@ -61,14 +61,14 @@ type WebhookNotification struct {
 	GroupLabels       map[string]string `json:"groupLabels"`
 	CommonLabels      map[string]string `json:"commonLabels"`
 	CommonAnnotations map[string]string `json:"commonAnnotations"`
-	ExternalUrl       string            `json:"externalURL"`
+	ExternalURL       string            `json:"externalURL"`
 	Alerts            []struct {
 		Status       string            `json:"status"`
 		Labels       map[string]string `json:"labels"`
 		Annotations  map[string]string `json:"annotations"`
 		StartsAt     string            `json:"startsAt"`
 		EndsAt       string            `json:"endsAt"`
-		GeneratorUrl string            `json:"generatorURL"`
+		GeneratorURL string            `json:"generatorURL"`
 	} `json:"alerts"`
 }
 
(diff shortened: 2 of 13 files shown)


src/github.com/matrix-org/go-neb/services/alertmanager/alertmanager.go
```diff
@@ -125,13 +125,14 @@ func (s *Service) Register(oldService types.Service, client *gomatrix.Client) er
...
-			if err != nil {
-				return fmt.Errorf("plain text template is invalid")
-			}
 		}
+
+		// validate the plain text template is valid
+		_, err := text.New("textTemplate").Parse(templates.TextTemplate)
+		if err != nil {
+			return fmt.Errorf("plain text template is invalid")
+		}
```
turt2live: This change is unrelated as far as I can tell.

Dockerfile
```diff
@@ -1,13 +1,18 @@
...
-RUN apk add --no-cache -t build-deps git gcc musl-dev go \
-    && go get -u github.com/constabulary/gb/... \
+RUN go get -u github.com/constabulary/gb/... \
+    && go get github.com/golang/lint/golint \
+    && go get github.com/fzipp/gocyclo \
     && gb vendor restore \
     && gb build -f github.com/matrix-org/go-neb
 
+# Ensures we're lint-free
+RUN /tmp/go-neb/hooks/pre-commit
```
turt2live: do we really need to lint in the dockerfile?

MTRNord: /me votes for no. As a pre-commit hook should ensure a lint error doesnt happen BEFORE it is even pushed. (Atleast thats the goal of it afaik)

michaelkaye: So from my use - I'm not a go developer and I don't have a local go development environment installed, so I've found that using docker to manage the local build and GOPATH and so on is very simple -- `docker build .` will always work and be reproducible locally and the same will happen in production.

So I just added the linter as part of the build, so it'll be run, as I'm not able to use the (optional, unenforced) pre-commit hook.

I'm happy to remove the linter from the docker build if that's needed, but I thought I'd explain why it was added, and I'll probably end up having to put it back in locally next time I develop on go-neb.

turt2live: In general it'd be nice to have the Docker images be for running the bot and less so for developing it, which means pulling in as little as possible. For the development environment problem, we could introduce a Vagrantfile or something that makes it a little less painful.

dajohi:
@michaelkaye @turt2live would love to see a go 1.11 update :)

Dockerfile
michaelkaye: So the docker image is still for running the bot - that's the non-build image that we copy to, which does not contain a linting step at runtime.

The change here is an update to the docker image for compiling the binary - so this linting is part of the build process, not the running process.

I believe linting should be considered part of a build process, failing linting should be FATAL, and we should have the same build process everywhere where possible, and so a 5s == 5% overhead on a build is not the end of the world, especially as go-neb has not had any code or PRs merged to it in over four months - I don't feel that 5s will really slow things down.

```
~/work/go-neb$ docker build --no-cache .
[+] Building 118.7s (14/14) FINISHED                                                                                                                                                                               
 => local://dockerfile (Dockerfile)                                                                                                                                                                           0.1s
 => => transferring dockerfile: 38B                                                                                                                                                                           0.0s
 => local://context (.dockerignore)                                                                                                                                                                           0.1s
 => => transferring context: 34B                                                                                                                                                                              0.0s
 => CACHED docker-image://docker.io/library/golang:1.11-alpine                                                                                                                                                0.0s
 => CACHED docker-image://docker.io/library/alpine:3.7                                                                                                                                                        0.0s
 => CACHED docker-image://docker.io/tonistiigi/copy:v0.1.3@sha256:e57a3b4d6240f55bac26b655d2cfb751f8b9412d6f7bb1f787e946391fb4b21b                                                                            0.0s
 => local://context                                                                                                                                                                                           0.2s
 => => transferring context: 6.80kB                                                                                                                                                                           0.0s
 => /bin/sh -c apk add --no-cache -t build-deps git gcc musl-dev go                                                                                                                                          18.6s
 => copy /src-0 tmp/go-neb                                                                                                                                                                                    1.2s
 => /bin/sh -c go get -u github.com/constabulary/gb/...     && go get github.com/golang/lint/golint     && go get github.com/fzipp/gocyclo     && gb vendor restore     && gb build -f github.com/matrix-or  89.7s
 => /bin/sh -c time /tmp/go-neb/hooks/pre-commit                                                                                                                                                              5.7s
 => copy /src-0/go-neb usr/local/bin/go-neb                                                                                                                                                                   0.8s
 => /bin/sh -c apk add --no-cache       ca-certificates       su-exec       s6                                                                                                                                1.4s
 => copy /src-0/root ./                                                                                                                                                                                       0.5s
 => exporting to image                                                                                                                                                                                        0.3s
 => => exporting layers                                                                                                                                                                                       0.3s
 => => writing image sha256:5220c05d313f533a8ac79c09c7d94317ab24556f69747281b899824d638f5b65                                                                                                                  0.0s

```

richvdh: +1, this looks sensible to me. @turt2live do you still have strong opinions against it?

turt2live: I have objections to running the linter in the container, but not enough to block this. 

The image should be built by CI, which should also be doing the linting for us. The layering in the image is to reduce the overall size, not to be our CI.

PR #254, opened 2018-11-26
Log alertmanager errors

michaelkaye:
When the template fails to compile, it should log the error to the alertmanager console to provide some feedback on the details of the failure.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,5 @@
 vendor/pkg
 vendor/src
 pkg
-hooks
 bin
 .git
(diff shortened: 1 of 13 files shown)


turt2live:
Can we split out the build process changes please?

turt2live:
I honestly have no idea what is going on here: is the change proposed supposed to be https://github.com/matrix-org/go-neb/pull/254/commits/e5ebd41c1392d321d6a9c4d414bea6ace3b42569 ?

richvdh:
I think this is superceded by #306

PR #265, opened 2019-02-15
Add mtgcard service

viccuad:
This service makes API calls to https://scryfall.com/docs/api/ and returns
information about "Magic: the Gathering" cards.

It provides a command:
!mtcard card name

And it provides expansion for:
[[card name]]

Both will return an html message with card information, links, and price.

I'm curious if you folks see it relevant to be added to go-neb or not.

If you want to see it live, hop to #magicthegathering:matrix.org (currently
running non-dedicated).


viccuad:
Closing in favour of #267 so I can iterate on other changes. Sorry for the noise.

PR #266, opened 2019-02-21
Make codebase pass the hook/pre-commit check

viccuad:


richvdh:
I think this is redundant to #253

PR #269, opened 2019-03-12
Improve formatting of RSS posts

thomas-profitt:
Screenshot_2019-03-11 Riot  2
Screenshot from 2019-03-11 19-06-47

Currently, if you have a room you use for several RSS feeds:

- It's difficult to quickly identify the RSS feed a given post came from, because the source is on the same line as all the other data, and it isn't clear where it ends, so you can't skim through them and identify what post came from what source by the length of the field.
- It's also not clear where the post title begins, for the same reason.
- Because the whole text of the link is included, in addition to the post title, it is liable to wrap to a 2nd line.
- Because of the " posted a new article", the already-long lines are made longer on both clients that support HTML-formatted Matrix messages and clients that don't. To convey that SomeRSSFeed is the source of a post, the context that the RSS bot is posting it, and a colon, as seen in the screenshots, should be enough.

I'm not familiar with Go, and I don't already have a test environment for running go-neb, so I didn't test this change. I'm hoping nothing is broken and that a quick review by a real Go developer and the CI will catch any problems I might've introduced.

### DIFF
--- a/src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
+++ b/src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
@@ -356,9 +356,8 @@ func itemToHTML(feed *gofeed.Feed, item gofeed.Item) gomatrix.HTMLMessage {
 		Format: "org.matrix.custom.html",
 		FormattedBody: fmt.Sprintf("<strong>%s</strong>:<br><a href=\"%s\"><strong>%s</strong></a>",
 			html.EscapeString(feed.Title), html.EscapeString(item.Link), html.EscapeString(item.Title)),
-			// SomeOne posted a new article: Title Of The Entry ( https://someurl.com/blag )
 			// FeedTitle:
-			// Title of the Entry which is a Link to It
+			// Title of the Entry, as a Link to It
 	  }
 }
 
--- a/src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
+++ b/src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
@@ -356,8 +356,9 @@ func itemToHTML(feed *gofeed.Feed, item gofeed.Item) gomatrix.HTMLMessage {
 		Format: "org.matrix.custom.html",
 		FormattedBody: fmt.Sprintf("<strong>%s</strong>:<br><a href=\"%s\"><strong>%s</strong></a>",
 			html.EscapeString(feed.Title), html.EscapeString(item.Link), html.EscapeString(item.Title)),
-			// FeedTitle:
-			// Title of the Entry, as a Link to It
+			// <strong>FeedTitle</strong>:
+			// <br>
+			// <a href="url-of-the-entry"><strong>Title of the Entry</strong></a>
 	  }
 }
 
--- a/src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
+++ b/src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
@@ -348,12 +348,18 @@ func (s *Service) sendToRooms(cli *gomatrix.Client, feedURL string, feed *gofeed
 	return nil
 }
 
-// SomeOne posted a new article: Title Of The Entry ( https://someurl.com/blag )
 func itemToHTML(feed *gofeed.Feed, item gofeed.Item) gomatrix.HTMLMessage {
-	return gomatrix.GetHTMLMessage("m.notice", fmt.Sprintf(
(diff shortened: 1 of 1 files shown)


src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
```diff
@@ -348,12 +348,17 @@ func (s *Service) sendToRooms(cli *gomatrix.Client, feedURL string, feed *gofeed
...
-	return gomatrix.GetHTMLMessage("m.notice", fmt.Sprintf(
-		"<i>%s</i> posted a new article: %s ( %s )",
-		html.EscapeString(feed.Title), html.EscapeString(item.Title), html.EscapeString(item.Link),
-	))
+	return &gomatrix.HTMLMessage{
+		Body: fmt.Sprintf("%s: %s (%s)",
+			html.EscapeString(feed.Title), html.EscapeString(item.Title), html.EscapeString(item.Link)),
+		MsgType: "m.notice",
+		Format: "org.matrix.custom.html",
+		FormattedBody: fmt.Sprintf("<strong>%s</strong>:<br><a href=\"%s\"><strong>%s</strong></a>",
```
thomas-profitt: An assumption I'm making here is that no URLs will have `"` in them, because I'm just interpolating.

It would be best to run the URL through a URL encoder that will escape the `"`, or even just substitute `"` with the proper URL escape code with crude string substitution, but I don't know what libraries I might use to do those things in Go, or exactly how to do it since I've never written a line of Go before.

anoadragon453: Does `html.EscapeString()` work for this purpose?

src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
```diff
@@ -348,12 +348,17 @@ func (s *Service) sendToRooms(cli *gomatrix.Client, feedURL string, feed *gofeed
...
+			html.EscapeString(feed.Title), html.EscapeString(item.Link), html.EscapeString(item.Title)),
+			// FeedTitle:
+			// Title of the Entry, as a Link to It
```
anoadragon453: What does this mean exactly? "and a Link to it"*?

thomas-profitt: Title of the entry, _as_ a link to it -- the title of the entry is a link to the entry (see screenshot from Riot)

src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
thomas-profitt: Looking at https://golang.org/pkg/html/#EscapeString , it looks like this will insert HTML escape codes into our URL, but we want URI escape codes.

Is there a Go REPL I can use to test this without having to set up and run my own go-neb?

thomas-profitt: I did some reading: HTML escape codes in HTML attribute values are indeed valid and are the proper way to escape text inside attribute values!

Testing the following code in https://play.golang.org, it looks to produce output valid by that:
```go
package main

import (
	"fmt"
	"html"
)

func main() {
	fmt.Printf("<a href=\"%s\">World</a>", html.EscapeString("https://example.com/<\"\">?this=that&that=this"))
}
```

src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
thomas-profitt: Changed in https://github.com/matrix-org/go-neb/pull/269/commits/b99b1b808fccddc7c2b2d54b01c845f133802e2a to clarify

src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
anoadragon453: Looks good.

P.S. Heads up you can share a play session so people can run the code already loaded in: https://play.golang.org/p/TgSLH8n-JyT

anoadragon453:
Don't mind the CI, it needs some updating.

bhush9:
I can't seem to build this with `rssbot.go:357:3: cannot use &gomatrix.HTMLMessage literal (type *gomatrix.HTMLMessage) as type gomatrix.HTMLMessage in return argument` ?

anoadragon453:
@bhush9 Thanks for reporting. Fixed in https://github.com/matrix-org/go-neb/pull/277

Issue #217, opened 2018-01-26
Limit size of gifs returned from giphy


turt2live:
44mb for a gif is enormous.

MurzNN:
Maybe better have option to send only url to image, instead of upload full image file to matrix? Most of Matrix clients will generate preview and play gif per click.

turt2live:
I like the inline gifs, but it is annoying to have a large gif posted that takes ages to download. Instead, the bot should pick a smaller gif to post.

PR #276, opened 2019-04-09
Add support to GIPHY bot for downsampled images

wpine215:
Addresses #217 , adding support for fixed_height_downsampled and fixed_width_downsampled images as per the GIPHY API to provide the option of reducing GIF file size further. Replaces use_downsized with image_type to support all four image types (original image, fixed height downsampled, fixed width downsampled, and downsized).

I'm relatively new to contributing to open source projects so if there is anything I could improve upon I am welcome to feedback, thank you!

---
Signed-off-by: William Pine <wpine215@gmail.com>

### DIFF
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -67,7 +67,7 @@ services:
     UserID: "@goneb:localhost" # requires a Syncing client
     Config:
       api_key: "qwg4672vsuyfsfe"
-      use_downsized: false
+      image_type: 0
 
   - ID: "guggy_service"
     Type: "guggy"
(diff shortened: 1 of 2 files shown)


config.sample.yaml
```diff
@@ -67,7 +67,7 @@ services:
     UserID: "@goneb:localhost" # requires a Syncing client
     Config:
       api_key: "qwg4672vsuyfsfe"
-      use_downsized: false
+      image_type: 0
```
Kegsay, line 70: I'd prefer string enums to integers as strings are more descriptive without knowing the mapping.

Issue #284, opened 2019-05-08
RSS: trac.torproject.org feeds don't have clickable links in Riot


Mikaela:
There is just nothing to click at Riot.

![Screenshot from 2019-05-09 00-37-40](https://user-images.githubusercontent.com/831184/57410129-841f5a80-71d9-11e9-807b-db37c31c761d.png)

```json
{
  "content": {
    "body": "Tor Bug Tracker &amp; Wiki: Ticket #23888: Creating a Snowflake WebExtension addon:  (https://trac.torproject.org/projects/tor/ticket/23888#comment:19)",
    "format": "org.matrix.custom.html",
    "formatted_body": "<strong>Tor Bug Tracker &amp; Wiki: Ticket #23888: Creating a Snowflake WebExtension addon</strong>:<br><a href=\"https://trac.torproject.org/projects/tor/ticket/23888#comment:19\"><strong></strong></a>",
    "msgtype": "m.notice"
  },
  "event_id": "$15573503543595335xtwTH:matrix.org",
  "origin_server_ts": 1557350354688,
  "sender": "@_neb_rssbot_=40mikaela=3aprivacytools.io:matrix.org",
  "type": "m.room.message",
  "unsigned": {
    "age": 921292
  },
  "room_id": "!zBMUYyApaSEJLVWozj:matrix.org"
}
```

How IRC (or Fruitiex/TeleIRC sees it):

`<RSSBot[mikaelapr> Tor Bug Tracker &amp; Wiki: Ticket #23888: Creating a Snowflake WebExtension addon:  (https://trac.torproject.org/projects/tor/ticket/23888#comment:19)`

Mikaela:
Another entry from that feed got a link, but it wasn't the full line as usually.

![Screenshot from 2019-05-09 00-45-26](https://user-images.githubusercontent.com/831184/57410577-abc2f280-71da-11e9-9e0a-344370e95c85.png)

```json
{
  "content": {
    "body": "Tor Bug Tracker &amp; Wiki: Ticket #23888: Creating a Snowflake WebExtension addon: cc changed (https://trac.torproject.org/projects/tor/ticket/23888#comment:20)",
    "format": "org.matrix.custom.html",
[...]

PR #285, opened 2019-05-08
Use feed's title if item does not have one

anoadragon453:
This prevents a link from not being shown if an item's title could not be extracted.

Workaround(?) fix for #284.

### DIFF
--- a/src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
+++ b/src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
@@ -350,8 +350,10 @@ func (s *Service) sendToRooms(cli *gomatrix.Client, feedURL string, feed *gofeed
 
 func itemToHTML(feed *gofeed.Feed, item gofeed.Item) gomatrix.HTMLMessage {
 	// If an item does not have a title, try using the feed's title instead
-	if item.Title == "" {
-		item.Title = feed.Title
+	// Create a new variable instead of mutating that which is passed in
+	itemTitle = item.Title
+	if itemTitle == "" {
+		itemTitle = feed.Title
 	}
 	
 	return gomatrix.HTMLMessage{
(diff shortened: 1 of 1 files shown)


src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
```diff
@@ -349,6 +349,11 @@ func (s *Service) sendToRooms(cli *gomatrix.Client, feedURL string, feed *gofeed
 }
 
 func itemToHTML(feed *gofeed.Feed, item gofeed.Item) gomatrix.HTMLMessage {
+	// If an item does not have a title, try using the feed's title instead
+	if item.Title == "" {
+		item.Title = feed.Title
```
jaywink: This is mutating the passed in object, which probably isn't bad in this case but in general I'd try to avoid it if possible as it can lead to hard to debug bugs. Possibly define a new variable `itemTitle` for example?

anoadragon453:
Edited, thanks :)

src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
```diff
@@ -349,13 +349,20 @@ func (s *Service) sendToRooms(cli *gomatrix.Client, feedURL string, feed *gofeed
 }
 
 func itemToHTML(feed *gofeed.Feed, item gofeed.Item) gomatrix.HTMLMessage {
+	// If an item does not have a title, try using the feed's title instead
+	// Create a new variable instead of mutating that which is passed in
+	itemTitle = item.Title
```
jaywink:
```
# github.com/matrix-org/go-neb/services/rssbot
rssbot.go:354:2: undefined: itemTitle
```

I think `itemTitle = item.Title`?

jaywink: Err, `itemTitle := item.Title`

jaywink: (also we should fix CI :D)

Kegsay:
Linting package no longer exists, fixed in #310 so merging for now.

src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
Kegsay: #310 fixes CI, I've added `:=`

PR #286, opened 2019-05-11
Update to go1.11 and fix linting

anoadragon453:
Hopefully should fix CI.


richvdh:
I think this is a dup of #253

PR #288, opened 2019-05-25
Add spaces between URL and parenthesis

vranki:
I'm getting complaints from users that the links created by RSS bot are not clickable as their client interprets the parenthesis as part of the URL. This change adds spaces between the URL and the parenthesis so that it looks sane and the URL's should be clickable on all clients.

Parenthesis can be part of URL in some cases, so it's understandable that clients consider them part of URL (although probably in wrong position).. Some discussion here: https://webmasters.stackexchange.com/questions/78110/is-it-bad-to-use-parentheses-in-a-url

I have not tested the change as it's trivial and I'm not running the bot myself.

Signed-off-by: Ville Ranki <ville.ranki@iki.fi>

### DIFF
--- a/src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
+++ b/src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
@@ -350,7 +350,7 @@ func (s *Service) sendToRooms(cli *gomatrix.Client, feedURL string, feed *gofeed
 
 func itemToHTML(feed *gofeed.Feed, item gofeed.Item) gomatrix.HTMLMessage {
 	return gomatrix.HTMLMessage{
-		Body: fmt.Sprintf("%s: %s (%s)",
+		Body: fmt.Sprintf("%s: %s ( %s )",
 			html.EscapeString(feed.Title), html.EscapeString(item.Title), html.EscapeString(item.Link)),
 		MsgType: "m.notice",
 		Format: "org.matrix.custom.html",


vranki:
Looks like travis ci is broken.

jaywink:
Thanks @vranki . I wonder if [using brackets](https://stackoverflow.com/questions/40568/are-square-brackets-permitted-in-urls) would be better than adding spaces, opinions?

Yeah CI is borked, #286 is trying to fix it.

jaywink:
Also please [sign-off your commit](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.rst#sign-off) as per the Matrix contributor guidelines, thanks!

vranki:
> I wonder if [using brackets](https://stackoverflow.com/questions/40568/are-square-brackets-permitted-in-urls) would be better than adding spaces, opinions?

We tested this, unfortunately Putty fails parsing and includes last ] in the URL. Adding spaces should solve the issue on all broken URL parsers. Parentheses could of course be replaced by brackets or even left out.

jaywink:
Right :+1: Good to go then, just need the sign-off for the commit as per contribution guideline for all matrix.org repositories.

vranki:
Signed off by editing the PR message.

jaywink:
Thanks!

jaywink:
@vranki matrix.org powered Go-NEB instance deployed with the fix

PR #292, opened 2019-06-27
Add safe-search option to google service

t3chguy:
Fixes https://github.com/matrix-org/go-neb/issues/291

Signed-off-by: Michael Telatynski 7t3chguy@gmail.com

### DIFF
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -81,6 +81,7 @@ services:
     Config:
       api_key: "AIzaSyA4FD39m9"
       cx: "AIASDFWSRRtrtr"
+      safe_search: false
 
   - ID: "imgur_service"
     Type: "imgur"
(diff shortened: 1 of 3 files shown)


turt2live:
can we do it per-room too? (for if the config flag is unset)

t3chguy:
/me has no idea how per-room stuff in go-neb works :D
I mean in theory you already can, by running one service per-room, using the same underlying API Key and Matrix Account, but perhaps that's not the cleanest

turt2live:
The github bot has a default repo config somewhere, so should be able to copy that

t3chguy:
image

t3chguy:
There, I couldn't find an actual image which safesearch filtered out but that could just be due to my quick google customsearch set up

Issue #264, opened 2019-02-15
Documentation/Examples for AlertManager Integration


lsnow2017:
I am struggling to connect my AlertManager install to the matrix bot. I have the following configs:

config.sample.yaml:
```
clients:
  - UserID: "@jeeves:matrix.mydomain.com"
    AccessToken: "<token>"
    HomeserverURL: "https://matrix.mydomain.com"
    Sync: true
    AutoJoinRooms: true
    DisplayName: "Jeeves"
services:  
   - ID: "prom_alerts"
     Type: "alertmanager"
     UserID: "@jeeves:matrix.mydomain.com"
     Config:
      webhook_url: "/prom_alerts"`
```

And for alert manager:
```
receiver: 'jeeves-webhook'
  routes:
  - match:
      service: logan-alert
    receiver: 'jeeves-webhook'

receivers:
- name: "jeeves-webhook"
  webhook_configs:
  - url: 'https://alerts.mydomain.com/services/hooks/cHJvbV9hbGVydHM'
    send_resolved: true
```

When the webhook fires I get this error from the bot output:
`INFO[0025] Failed to load service                        error=sql: no rows in result set service_id=prom_alerts`

Some additional documentation/any guidance would be great. Thank you!

jaywink:
Did you ever manage to get this to work? Certainly can't see anything wrong with your configuration, at least concerning the error raised :thinking:

maralorn:
@jaywink We can reproduce this bug.
```
level=info msg="Incoming webhook request" path=/services/hooks/YWxlcnRtYW5hZ2VyX3NlcnZpY2UK
level=info msg="Failed to load service" error="sql: no rows in result set" service_id="alertmanager_service\n"
```
So apparently in the moment when the webhook is triggered something in the bot tries to get something from the database, which shouldn‘t happen, when a config file is given.

maralorn:
Sorry, for the noise. The issue for us was the wrong webhook url in the #318 . Please merge that one. I can confirm it works!

mweinelt:
The base64 string in the request and the example config is wrong, see #318.

jaywink:
Closing this as OP has not come back with more details and #318 is merged, thanks all.

PR #294, opened 2019-07-10
Add an example Alertmanager config

jaywink:
For when running in config file mode. The same example can be used
to manually insert into the SQLite database or figure out
how to use the services API to configure Alertmanager service

Refs: #264
Signed-off-by: Jason Robinson <jasonr@matrix.org>


jaywink:
I'd love to also get `gh-pages` updated to get a link to alertmanager service example in the readme.

PR #300, opened 2019-08-16
Try to fix RSS

liberodark:
Fix https://github.com/matrix-org/go-neb/issues/299

### DIFF
--- a/src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
+++ b/src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
@@ -351,11 +351,11 @@ func (s *Service) sendToRooms(cli *gomatrix.Client, feedURL string, feed *gofeed
 func itemToHTML(feed *gofeed.Feed, item gofeed.Item) gomatrix.HTMLMessage {
 	return gomatrix.HTMLMessage{
 		Body: fmt.Sprintf("%s: %s ( %s )",
-			html.EscapeString(feed.Title), html.EscapeString(item.Title), html.EscapeString(item.Link)),
+			html.UnescapeString(feed.Title), html.UnescapeString(item.Title), html.UnescapeString(item.Link)),
 		MsgType: "m.notice",
 		Format: "org.matrix.custom.html",
 		FormattedBody: fmt.Sprintf("<strong>%s</strong>:<br><a href=\"%s\"><strong>%s</strong></a>",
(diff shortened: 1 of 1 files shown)


liberodark:
What is this bug onyour Travis : 

`../../../golang.org/x/tools/go/internal/gcimporter/bexport.go:212: obj.IsAlias undefined (type *types.TypeName has no field or method IsAlias)`

turt2live:
No idea what that bug is, I think it can be ignored.

liberodark:
> No idea what that bug is, I think it can be ignored.

For what ignored ? im try to fix that for more compatility with websites.

Best Regards

src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
```diff
@@ -351,11 +351,11 @@ func (s *Service) sendToRooms(cli *gomatrix.Client, feedURL string, feed *gofeed
 func itemToHTML(feed *gofeed.Feed, item gofeed.Item) gomatrix.HTMLMessage {
 	return gomatrix.HTMLMessage{
 		Body: fmt.Sprintf("%s: %s ( %s )",
-			html.EscapeString(feed.Title), html.EscapeString(item.Title), html.EscapeString(item.Link)),
+			html.UnescapeString(feed.Title), html.UnescapeString(item.Title), html.UnescapeString(item.Link)),
 		MsgType: "m.notice",
 		Format: "org.matrix.custom.html",
 		FormattedBody: fmt.Sprintf("<strong>%s</strong>:<br><a href=\"%s\"><strong>%s</strong></a>",
-			html.EscapeString(feed.Title), html.EscapeString(item.Link), html.EscapeString(item.Title)),
+			html.UnescapeString(feed.Title), html.UnescapeString(item.Link), html.UnescapeString(item.Title)),
```
Kegsay, line 358: This allows XSS attacks if the title contains stuff like `"&#60;script&#62;stuff"`

Kegsay:
RSS feeds all vary in what their items contain. Sometimes they are HTML, sometimes plaintext, sometimes escaped and sometimes not. We apply best-effort heuristics to try to support the widest range of feeds possible, but we will inevitably miss some.

This PR also introduces a security vulnerability.

On this basis, I'm closing this PR for now. It would be helpful if you could add an example test RSS feed which exhibits this particular failure mode so we can investigate this further.

Issue #298, opened 2019-07-29
RSS: tell the feed author if one exists


Mikaela:
If the feed has `<author>Name</author>`, I think that should be told in the RSS announcement messages somehow. This would be especially interesting in cases where there are multiple authors in the same feed. I think English makes this easy and Go Neb could simply add "by AuthorName" to the end of the titles like [bitbot](https://github.com/jesopo/bitbot) does:

Some examples (emphasises mine):

> write.privacytools.io Reader:
> Why you should start your own blog?

* https://write.privacytools.io/read/feed/ (WriteAs instance) through Go Neb which has `<author>Goo</author>` for this entry.

> [RSS] Private Internet Access Blog: What can we learn from Palantir’s secret user manual for police users of its big data software? *by Glyn Moody* - https://www.privateinternetaccess.com/blog/2019/07/what-can-we-learn-from-palantirs-secret-user-manual-for-police-users-of-its-big-data-software/

* https://www.privateinternetaccess.com/blog/feed/ through [bitbot](https://github.com/jesopo/bitbot)

> [RSS] Ubuntu security notices: USN-4077-1: tmpreaper vulnerability by Ubuntu Security Team - https://usn.ubuntu.com/4077-1/

* https://usn.ubuntu.com/rss.xml which doesn't have author tags through [bitbot](https://github.com/jesopo/bitbot)

> [RSS] The Reykjavik Grapevine: Icelanders Suffering Through A Weather Mildwave *by Sveinbjörn Pálsson* - https://grapevine.is/news/2019/07/29/icelanders-suffering-through-a-weather-mildwave/

* https://grapevine.is/feed/ through [bitbot](https://github.com/jesopo)

Issue #296, opened 2019-07-16
RSS filter options


turt2live:
To filter out releases from a github feed the room doesn't care about

lub:
Kinda related to #236 I guess

Mikaela:
I guess https://github.com/matrix-org/go-neb/issues/298 is also partially related in case the RSS filtering options will include filtering by article author.

PR #311, opened 2020-03-13
RSS show author & ability to filter items

### LINKED ISSUE #236: RSS filter keywords
It would be nice to have the option to filter for user supplied keywords (must include or must not include) of a specific RSS feed.

e.g. to filter for specific videos of a youtube channel feed
or filter for "changelog" in a more general product blog

nikofil:
closes #236, #296, #298

### DIFF
--- a/src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
+++ b/src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
@@ -327,6 +327,10 @@ func (s *Service) newItems(feedURL string, allItems []*gofeed.Item) (items []gof
 		//   This will inevitably break for some people, but that group of people are probably smaller, so *shrug*.
 		i.Title = html.UnescapeString(i.Title)
 		i.Description = html.UnescapeString(i.Description)
+		if i.Author != nil {
+			i.Author.Name = html.UnescapeString(i.Author.Name)
+			i.Author.Email = html.UnescapeString(i.Author.Email)
+		}
 
 		items = append(items, *i)
 	}
@@ -355,18 +359,30 @@ func itemToHTML(feed *gofeed.Feed, item gofeed.Item) gomatrix.HTMLMessage {
 	if itemTitle == "" {
 		itemTitle = feed.Title
 	}
-	
+
+	fmtBody := fmt.Sprintf("<strong>%s</strong>:<br><a href=\"%s\"><strong>%s</strong></a>",
+		html.EscapeString(feed.Title), html.EscapeString(item.Link), html.EscapeString(itemTitle))
+	if item.Author != nil {
+		if len(item.Author.Name) > 0 && len(item.Author.Email) > 0 {
+			fmtBody += fmt.Sprintf(" by <a href=\"mailto:%s\">%s</a>", html.EscapeString(item.Author.Email),
+				html.EscapeString(item.Author.Name))
+		} else if len(item.Author.Name) > 0 {
+			fmtBody += fmt.Sprintf(" by %s", html.EscapeString(item.Author.Name))
+		} else if len(item.Author.Email) > 0 {
+			fmtBody += fmt.Sprintf(" by <a href=\"mailto:%s\">%s</a>", html.EscapeString(item.Author.Email),
+				html.EscapeString(item.Author.Email))
+		}
+	}
 	return gomatrix.HTMLMessage{
 		Body: fmt.Sprintf("%s: %s ( %s )",
-			html.EscapeString(feed.Title), html.EscapeString(item.Title), html.EscapeString(item.Link)),
-		MsgType: "m.notice",
-		Format: "org.matrix.custom.html",
-		FormattedBody: fmt.Sprintf("<strong>%s</strong>:<br><a href=\"%s\"><strong>%s</strong></a>",
-			html.EscapeString(feed.Title), html.EscapeString(item.Link), html.EscapeString(itemTitle)),
-			// <strong>FeedTitle</strong>:
-			// <br>
-			// <a href="url-of-the-entry"><strong>Title of the Entry</strong></a>
-	  }
+			html.EscapeString(feed.Title), html.EscapeString(itemTitle), html.EscapeString(item.Link)),
+		MsgType:       "m.notice",
+		Format:        "org.matrix.custom.html",
+		FormattedBody: fmtBody,
+		// <strong>FeedTitle</strong>:
+		// <br>
+		// <a href="url-of-the-entry"><strong>Title of the Entry</strong></a>
+	}
 }
 
 func ensureItemsHaveGUIDs(feed *gofeed.Feed) {
--- a/src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
+++ b/src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
@@ -7,7 +7,9 @@ import (
 	"html"
 	"net/http"
 	"strconv"
+	"strings"
 	"time"
+	"unicode"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/die-net/lrucache"
@@ -33,6 +35,30 @@ var (
 )
 
 const minPollingIntervalSeconds = 60 * 5 // 5 min (News feeds can be genuinely spammy)
+
+// includeRules contains the rules for including or excluding a feed item. For the fields Author, Title
+// and Description in a feed item, there can be some words specified in the config that determine whether
+// the item will be displayed or not, depending on whether these words are included in that field.
+//
+//   - If specified in the `must_include` field, the feed item must include at least one word for each field
+//     that has been specified. This means that if some words have been specified for both Author and Title,
+//     both the Author and Title must contain at least one of their respective words or the item will be skipped.
+//   - If specified in the `must_not_include` field, the feed item fields must not contain any of the words
+//     that were specified for each field. This means that if some words have been specified for both Author
+//     and Title, if either of them includes at least one of their respective words, the item will be skipped,
+//     even in the case that the item matched the `must_include` rules.
+//
+//   In both cases, specifying an empty list for a field or not specifying anything causes the field to be ignored.
+//   The field being checked each time will be split into words (any non-alphanumeric character starts a new word)
+//   and they will be checked against the provided list.
+type includeRules struct {
+	// Author is a case-sensitive list of words that the author name must contain or not contain.
+	Author []string `json:"author"`
+	// Title is a case-sensitive list of words that the author name must contain or not contain.
+	Title []string `json:"title"`
+	// Description is a case-sensitive list of words that the author name must contain or not contain.
+	Description []string `json:"description"`
+}
 
 // Service contains the Config fields for this service.
 //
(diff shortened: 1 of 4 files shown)


src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
```diff
@@ -34,6 +35,16 @@ var (
 
 const minPollingIntervalSeconds = 60 * 5 // 5 min (News feeds can be genuinely spammy)
 
+// includeRules contains the rules for including or excluding a feed item.
+type includeRules struct {
+	// Author is a list of strings that the author name must match or not match.
```
Kegsay, line 40: Please state whether there is case sensitivity.

src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
```diff
@@ -302,7 +317,41 @@ func (s *Service) queryFeed(feedURL string) (*gofeed.Feed, []gofeed.Item, error)
 	return feed, items, nil
 }
 
+func containsAny(item string, list []string) bool {
+	for _, i := range list {
+		if strings.Contains(item, i) {
+			fmt.Println("contains", item, i)
```
Kegsay, line 323: Spurious logging

src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
```diff
@@ -34,6 +35,16 @@ var (
 
 const minPollingIntervalSeconds = 60 * 5 // 5 min (News feeds can be genuinely spammy)
 
+// includeRules contains the rules for including or excluding a feed item.
+type includeRules struct {
+	// Author is a list of strings that the author name must match or not match.
+	Author []string `json:"author"`
```
Kegsay, line 41: Also please state what happens in the empty set case of`[]`. If I have `include: { author: [] }` does that filter everything out which has an author tag? We likely will want to treat empty values as not-set, so no filtering is applied.

src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
```diff
@@ -34,6 +35,16 @@ var (
 
 const minPollingIntervalSeconds = 60 * 5 // 5 min (News feeds can be genuinely spammy)
 
+// includeRules contains the rules for including or excluding a feed item.
+type includeRules struct {
```
Kegsay, line 39: You also need to state how include/exclude clashes are resolved. If I include author "foo" and exclude title "bar" and an article has both author Foo and title Bar, does it get included or excluded?

src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
```diff
@@ -302,7 +317,41 @@ func (s *Service) queryFeed(feedURL string) (*gofeed.Feed, []gofeed.Item, error)
...
+			return true
+		}
+	}
+	return false
+}
+
+func itemFiltered(i *gofeed.Item, mustInclude, mustNotInclude *includeRules) bool {
+	// If `must_include` terms have been specified, the feed item must include at least one item for each field
+	// that has been specified. This means that if both an Author and Title substring have been specified,
+	// both the Author and Title must contain these substrings or the item will be skipped.
```
Kegsay, line 333: So basically you're ANDing the terms instead of ORing them, which makes sense. This should probably be situated all in one place (I would propose `includeRules` struct) so the reader can fully see the algorithm you wish to apply, without having to hunt down random functions in this file.

src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
```diff
@@ -302,7 +317,41 @@ func (s *Service) queryFeed(feedURL string) (*gofeed.Feed, []gofeed.Item, error)
...
+		containsAny(i.Description, mustNotInclude.Description) {
+		return true
+	}
+	return false
+}
+
 func (s *Service) newItems(feedURL string, allItems []*gofeed.Item) (items []gofeed.Item) {
+	mustInclude := s.Feeds[feedURL].MustInclude
+	mustNotInclude := s.Feeds[feedURL].MustNotInclude
+	fmt.Println(mustInclude, len(mustInclude.Description))
```
Kegsay, line 353: Spurious logging

src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
```diff
@@ -327,8 +376,14 @@ func (s *Service) newItems(feedURL string, allItems []*gofeed.Item) (items []gof
 		//   This will inevitably break for some people, but that group of people are probably smaller, so *shrug*.
 		i.Title = html.UnescapeString(i.Title)
 		i.Description = html.UnescapeString(i.Description)
+		if i.Author != nil {
+			i.Author.Name = html.UnescapeString(i.Author.Name)
+			i.Author.Email = html.UnescapeString(i.Author.Email)
+		}
 
-		items = append(items, *i)
+		if !itemFiltered(i, &mustInclude, &mustNotInclude) {
```
Kegsay, line 384: Why are you passing pointers to this function? It seems odd since we don't do it anywhere else. You can even just inline the whole lot and do:
```
if !itemFiltered(i, s.Feeds[feedURL].MustInclude, s.Feeds[feedURL].MustNotInclude) {
    ...
}
```

which reduces the number of local vars hanging around.

src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
```diff
@@ -349,17 +404,26 @@ func (s *Service) sendToRooms(cli *gomatrix.Client, feedURL string, feed *gofeed
 }
 
 func itemToHTML(feed *gofeed.Feed, item gofeed.Item) gomatrix.HTMLMessage {
+	body := fmt.Sprintf("<strong>%s</strong>:<br><a href=\"%s\"><strong>%s</strong></a>",
+		html.EscapeString(feed.Title), html.EscapeString(item.Link), html.EscapeString(item.Title))
+	if item.Author != nil {
+		if len(item.Author.Email) > 0 {
+			body += fmt.Sprintf(" by <a href=\"mailto:%s\">%s</a>", html.EscapeString(item.Author.Email),
+				html.EscapeString(item.Author.Name))
```
Kegsay, line 412: If the author has an email but no name this will be a problem. Should probably default to the email in that case.

src/github.com/matrix-org/go-neb/services/rssbot/rssbot_test.go
```diff
@@ -103,3 +116,55 @@ func TestHTMLEntities(t *testing.T) {
...
+	}
+
+	// Recreate rssbot client
+	rssbot, err = createRSSClient(feedURL)
+	if err != nil {
+		t.Fatal("Failed to create RSS bot: ", err)
+	}
+
+	feed = rssbot.Feeds[feedURL]
+	feed.MustNotInclude.Author = []string{"kid"}
```
Kegsay, line 162: That's an interesting edge case. I'm not sure we want to do that. We probably want to treat each item as a word rather than a partial string so the filter "Jon" catches "Jon Snow" but not "Jonathan Creek".

src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
nikofil: To avoid copying the two `includeRules` structs each time, though they're not that big. As for the vars, I declare them to avoid doing a map lookup for each item.c

src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
nikofil: Yup, I forgot a couple of these. Removed now.

src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
nikofil: Done!

src/github.com/matrix-org/go-neb/services/rssbot/rssbot_test.go
nikofil: I see. Changed the code to split the string into words (delimited by non-alphanumeric) and the documentation to reflect that.

src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
nikofil: I put some more descriptive comments in all places (but mostly on `includeRules`). Let me know if you think something is not explained well enough.

src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
nikofil: Indeed, that's what already happened (empty == unset). Added this to the docs.

src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
nikofil: Yes, it's case sensitive atm. Added to the docs - let me know if you think case-insensitive is better.

src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
nikofil: Added to the docs.

src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
Kegsay: SGTM

src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
```diff
@@ -349,17 +422,29 @@ func (s *Service) sendToRooms(cli *gomatrix.Client, feedURL string, feed *gofeed
 }
 
 func itemToHTML(feed *gofeed.Feed, item gofeed.Item) gomatrix.HTMLMessage {
+	body := fmt.Sprintf("<strong>%s</strong>:<br><a href=\"%s\"><strong>%s</strong></a>",
+		html.EscapeString(feed.Title), html.EscapeString(item.Link), html.EscapeString(item.Title))
```
Kegsay, line 426: Please incorporate https://github.com/matrix-org/go-neb/pull/285/files

src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
Kegsay: Let's see how many people complain with it being case-sensitive :)

Kegsay:
The merge conflict is due to #285 - please fix that and LGTM!

src/github.com/matrix-org/go-neb/services/rssbot/rssbot.go
nikofil: Should https://github.com/matrix-org/go-neb/pull/311/files/97db7ae6dc21d03c11c430bd4aca2da5d233d016#diff-ae310978bc4c449759bda6db50889e71R440 also be using `itemTitle` instead of `item.Title`?

Kegsay: Yes please.

nikofil: done!

PR #322, opened 2020-06-06
[WIP] Switch Gomatrix for Mautrix

nikofil:
This PR switches all uses of Gomatrix in go-neb for Mautrix as a first step towards e2ee. The changes are mainly using typed `RoomIDs` / `UserIDs` instead of simple strings and replacing the use of `TextMessage` / `ImageMessage` etc. with `MessageEventContent`s.

This currently will not work for the reason that the default syncer of Mautrix does not attach the room ID to the event message as Gomatrix does, so there is no way for handlers to get the room ID.

This should be fixed with a PR to Mautrix which will be done in the future. For this reason, this is marked as a WIP, however any other comments are welcome. If you wish to test it you can simply add a line `event.RoomID = "!someRoom"` to the start of `onMessageEvent`.

Other than that, everything seems to work. I've tried all services except for Jira (couldn't figure how to set up as session), Slack and Travis (these 2 were simple enough that I didn't try) and Guggy (site is down?).


services/giphy/giphy.go
```diff
@@ -130,16 +133,16 @@ func (s *Service) searchGiphy(query string) (*result, error) {
 	return &search.Data, nil
 }
 
-func asInt(strInt string) uint {
-	u64, err := strconv.ParseUint(strInt, 10, 32)
+func asInt(strInt string) int {
+	u64, err := strconv.ParseInt(strInt, 10, 32)
```
Kegsay, line 137: Why did you change this?

services/utils/utils_test.go
```diff
@@ -0,0 +1,19 @@
...
+	mevt "maunium.net/go/mautrix/event"
+)
+
+func TestHTMLStrip(t *testing.T) {
+	msg := `before <hello a="b"><inside />during</hello> after`
+	stripped := StrippedHTMLMessage(mevt.MsgNotice, msg)
+	if stripped.MsgType != mevt.MsgNotice {
+		t.Fatalf("Expected MsgType %v, got %v", mevt.MsgNotice, stripped.MsgType)
+	}
+	expected := "before during after"
```
Kegsay, line 15: It would be worth having some encoded chars in there to make sure `UnescapeString` is working correctly. E.g `&lt;`

services/giphy/giphy.go
nikofil: Because all of `FileInfo`s fields are ints in mautrix (and not uints as in gomatrix).

clients/clients.go
```diff
@@ -299,15 +310,14 @@ func (c *Clients) onBotOptionsEvent(client *gomatrix.Client, event *gomatrix.Eve
...
-		return // not our member event
-	}
-	m := event.Content["membership"]
-	membership, ok := m.(string)
-	if !ok {
+func (c *Clients) onRoomMemberEvent(client *mautrix.Client, event *mevt.Event) {
+	if err := event.Content.ParseRaw(mevt.StateMember); err != nil {
 		return
 	}
+	if *event.StateKey != client.UserID.String() {
```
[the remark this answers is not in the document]
nikofil: While I didn't write this line, I understand from https://matrix.org/docs/spec/client_server/r0.6.1#state-event-fields that a state key should always exist. I'll add a nil check to be sure if you think that it could be nil.

services/utils/utils_test.go
nikofil: Good point, added

nikofil:
Fixed the build configs so that Travis builds.

clients/clients.go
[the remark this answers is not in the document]
Kegsay: It can be `nil`, nothing stops anyone from sending an `m.room.member` event without a state key.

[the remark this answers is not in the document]
nikofil: done :) Also added a similar nil check to `onBotOptionsEvent`.

electricOzone:
Am I understanding this right? "This currently will not work"

Does this mean that go-neb is broken (in master) until mautrix is updated?

electricOzone:
I just rolled my docker-compose back to matrixdotorg/go-neb:anoadragon453-patch-1, and go-neb seems to be working again on my homeserver. Why was this merged to master when there was a known conflict with mautrix? This should be rolled back ASAP.

nikofil:
Yes, you're right. PR #324 would also fix this, or I could prepare a smaller PR until that can be reviewed to fix this. Or it could just be reverted for now.

Kegsay:
I somehow entirely missed this. We'll land #324 to fix this for now.

Kegsay:
Filed the above issue.

PR #327, opened 2020-07-02
Integration tests

### LINKED ISSUE #326: Add integration tests
I saw green and happiness in https://github.com/matrix-org/go-neb/pull/322 but in reality it was very badly broken. Whilst we do have tests for services, we don't have testing around the bot framework itself. We should really add this to prevent humans making mistakes.

Specifically, a fake homeserver for Go-NEB to send `/sync` requests to as well as configuring via the HTTP API e.g `/configureClient` should do it. We're not interested in testing any specific service, so just the echo service will do.

nikofil:
* Add integration tests for neb joining an unencrypted / encrypted room and sending a message

(closes #326)

### DIFF
--- a/goneb_services_test.go
+++ b/goneb_services_test.go
@@ -16,49 +16,7 @@ import (
 	mevt "maunium.net/go/mautrix/event"
 )
 
-func TestConfigureClient(t *testing.T) {
-	mux := http.NewServeMux()
-	mxTripper := newMatrixTripper()
-	setup(envVars{
-		BaseURL:      "http://go.neb",
-		DatabaseType: "sqlite3",
-		DatabaseURL:  ":memory:",
-	}, mux, &http.Client{
-		Transport: mxTripper,
-	})
-
-	mxTripper.ClearHandlers()
-	mockWriter := httptest.NewRecorder()
-	syncChan := make(chan string)
-	mxTripper.HandlePOSTFilter("@link:hyrule")
-	mxTripper.Handle("GET", "/_matrix/client/r0/sync",
-		func(req *http.Request) (*http.Response, error) {
-			syncChan <- "sync"
-			return newResponse(200, `{
-				"next_batch":"11_22_33_44",
-				"rooms": {}
-			}`), nil
-		},
-	)
-
-	mockReq, _ := http.NewRequest("POST", "http://go.neb/admin/configureClient", bytes.NewBufferString(`
-	{
-		"UserID":"@link:hyrule",
-		"HomeserverURL":"http://hyrule.loz",
-		"AccessToken":"dangeroustogoalone",
-		"Sync":true,
-		"AutoJoinRooms":true
-	}`))
-	mux.ServeHTTP(mockWriter, mockReq)
-	expectCode := 200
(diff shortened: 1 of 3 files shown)


goneb_services_test.go
```diff
@@ -54,3 +57,310 @@ func TestConfigureClient(t *testing.T) {
...
+						"sender_key":"%s",
+						"ciphertext":"%s",
+						"session_id":"%s"
+					},
+					"origin_server_ts": 10000,
+					"unsigned": {"age": 100},
+					"event_id": "evt125"
+				}]}}}
+		}
+	}`, identityKeyMock, string(ciphertext), ogsMock.ID())
```
Kegsay, line 357: Careful doing this as a JSON string, are we sure these 3 values only ever contain alphanumerics (no newlines or special chars?)

goneb_services_test.go
```diff
@@ -54,3 +57,310 @@ func TestConfigureClient(t *testing.T) {
...
+		"Type": "echo",
+		"Id": "test_echo_service",
+		"UserID": "@link:hyrule",
+		"Config": {}
+	}`))
+	mux.ServeHTTP(mockWriter, serviceConfigReq)
+
+	// get the initial syncs out of the way
+	reqChan <- `{"next_batch":"11_22_33_44", "rooms": {}}`
+	reqChan <- `{"next_batch":"11_22_33_44", "rooms": {}}`
```
Kegsay, line 124: It's unclear why you do this twice. Perhaps a `reqChan` per user (echo service and main bot I guess is why this is twice?).

nikofil:
Travis did run it: https://travis-ci.org/github/matrix-org/go-neb/builds/704347926
For some reason the link to Travis stops showing on Github after a while?

goneb_services_test.go
nikofil: I believe first one is because of https://github.com/matrix-org/go-neb/blob/1be6c8c1cd445b9d1bbfcf955afc15b6c4da4c27/clients/bot_client.go#L124 which is done to get the state for all rooms which might be missing otherwise.
Second one is due to https://github.com/tulir/mautrix-go/blob/a45738638b0d43b9f1e7284dabeae26a117fb1e3/sync.go#L255 : when we're actually in the sync loop syncs without a `since` are ignored. As the first sync above does not set the syncer's `next_batch` var (as it's done outside the syncer's loop) we need a second one that will set the `next_batch` so that the next ones will be processed.

goneb_services_test.go
nikofil: Yes, they're all base64.

goneb_services_test.go
Kegsay: That's pretty fragile then. Perhaps it would be better to automatically send back an empty sync response if `since` is missing in the request? Then you wouldn't need this at all.

nikofil: True, that cleans up things a bit. I also moved the common test prologue which was setting up the mock server to a new func in a new commit.

Kegsay:
Please don't force push all the time as it makes it hard to see the diff from before.

PR #330, opened 2020-07-10
Docs about device ID & update mautrix version

### LINKED ISSUE #328: Device ID is not set which will result in E2E encryption/decryption not working
so I did a git pull from master etc to run go-neb with out the proxy since I notice you guys have closed the ticket to enable e2ee as done.
but I get this when loggin in:
```
 Device ID is not set which will result in E2E encryption/decryption not working 
```
where do I add this to the config? seems the wiki has not been updated to any new settings?

Thanks

nikofil:
* Add some docs about how to obtain and use the device ID
* Remove unnecessary ParseRaw calls
* Add `GetEncryptionEvent` for retrieving the room's encryption params
* Fix tests for new mautrix changes
* Update mautrix version to 0.6.0

(closes #328)


r3k2:
thanks! when is this going to be push into master?

PR #346, opened 2021-03-01
Respect rate limits by not continuing to send if 429 received.

michaelkaye:
I'm aware this will drop messages on the floor - but given we do not retry these messages that are returning 429 errors anyway, there's little point in continuing to retry and it only adds load to the live server.

### DIFF
--- a/services/rssbot/rssbot.go
+++ b/services/rssbot/rssbot.go
@@ -215,7 +215,10 @@ func (s *Service) OnPoll(cli types.MatrixClient) time.Time {
 					"feed_url":   u,
 					log.ErrorKey: err,
 					"item":       item,
-				}).Error("Failed to send item to room")
+				}).Error("Failed to send item to room due to 429; aborting further sends")
+				// no point continuing if we errored due to a 429 - we'll just hit the rate limit
+				// again and again
+				break
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


services/rssbot/rssbot.go
```diff
@@ -415,7 +418,11 @@ func (s *Service) sendToRooms(cli types.MatrixClient, feedURL string, feed *gofe
 	logger.Info("Sending new feed item")
 	for _, roomID := range s.Feeds[feedURL].Rooms {
 		if _, err := cli.SendMessageEvent(roomID, mevt.EventMessage, itemToHTML(feed, item)); err != nil {
-			logger.WithError(err).WithField("room_id", roomID).Error("Failed to send to room")
+			if err.Code == 429  {
```
tulir, line 421: I think it should be more like this, the Go `error` interface doesn't have a `.Code` field:

```suggestion
			if httpErr := err.(mautrix.HTTPError); httpErr.IsStatus(429) {
```

(may require updating mautrix-go)

tulir: Or with the fancy modern [`errors.As()`](https://golang.org/pkg/errors/#As) function

```suggestion
			if httpErr := mautrix.HTTPError{}; errors.As(err, &httpErr) && httpErr.IsStatus(429)  {
```

Kegsay:
CI is failing and it would be nice to add a test to `rssbot_test.go`.

michaelkaye:
Happy for someone to take this on, very much don't wait for me - i just want it not to chew CPU cycles on matrix.org with 429 errors.

I might find some time in a few months to learn enough go, as it seems the library upgrade wasn't as simple as advertised.

Kegsay:
I'll see if I can get to this this week.

PR #351, opened 2021-07-18
upgrade mautrix to 0.9.15

Mic92:
fixes https://github.com/matrix-org/go-neb/issues/347

PR #350, opened 2021-05-19
Go 1.16 support and some general improvements

daenney:
* Move go-neb to be built by 1.16 and update the Dockerfile to use newer bases
  Just for general house keeping purposes and getting access to new features in go
* Replace the deprecated golint with staticcheck and fix the linter warnings
  Golint has been [deprecated and frozen](https://github.com/golang/go/issues/38968)
* Fix the sqlite3 package version which also resolves a build issue on newer gcc's
  The v2 tags were erroneous, move us back to v1 tags
* Migrate CI from Travis to Github Actions
  We've started to use GHA for more things on matrix-org, this moves neb in the same direction. The one thing we lose now is the Travis-CI webhook notifications, but given GHA will poke the author of the PR if their builds fail that feels like an acceptable trade-off. You can see what a build ends up looking like [over here](https://github.com/daenney/go-neb/runs/2618457592?check_suite_focus=true)
* Upgrade to the current release of mautrix-go
  We need this in order to be able to build with newer versions of Go

PR #355, opened 2021-07-21
CI: Ensure we run on go.mod|sum changes

daenney:
I noticed that after the rebase for #351 CI didn't run, whereas it should (since we're upgrading packages). This happens because as part of #350 we forgot to include go.mod|sum files as triggers.


Half-Shot (approved):
Doesn't look dangerous, have a tick

PR #357, opened 2021-08-02
Add MinimumPowerLevel option to clients

Half-Shot:
This change adds an option to specify the minimum PL level of the inviting user for the bot to remain joined to the room. This change will only affect services which accept invites to rooms, so services like GitHub need a different change.

### DIFF
--- a/clients/clients.go
+++ b/clients/clients.go
@@ -303,7 +303,7 @@ func (c *Clients) onBotOptionsEvent(client *mautrix.Client, event *mevt.Event) {
 	}
 }
 
-func (c *Clients) onRoomMemberEvent(client *mautrix.Client, event *mevt.Event) {
+func (c *Clients) onRoomMemberEvent(client *BotClient, event *mevt.Event) {
 	if event.StateKey == nil || *event.StateKey != client.UserID.String() {
 		return // not our member event
 	}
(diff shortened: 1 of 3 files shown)


clients/clients.go
```diff
@@ -316,14 +316,35 @@ func (c *Clients) onRoomMemberEvent(client *mautrix.Client, event *mevt.Event) {
...
+		if client.config.MinimumPowerLevel == 0 {
+			return
+		}
+		if err := client.StateEvent(rID, mevt.StatePowerLevels, "", &plContent); err != nil {
+			logger.WithError(err).Print("Failed to get powerlevels from room")
+			client.LeaveRoom(rID)
+			return
+		}
+		var pl = plContent.Users[event.Sender.String()]
+		if pl == 0 {
```
jaywink, line 342: Technically this is wrong - the user could have a PL of 0 set explicitly, and afaict in that case the PL should not revert to `UsersDefault`. It should revert to `UsersDefault` if the user has not explicitly been set any power level.

However I'm not sure in this context does `plContent.Users` have all users or just those who have a PL set? If the latter, wont this whole thing crash since there will be nothing in the map for that user?

clients/clients.go
```diff
@@ -372,7 +393,7 @@ func (c *Clients) initClient(botClient *BotClient) error {
 
 	if config.AutoJoinRooms {
 		syncer.OnEventType(mevt.StateMember, func(_ mautrix.EventSource, event *mevt.Event) {
-			c.onRoomMemberEvent(client, event)
+			c.onRoomMemberEvent(botClient, event)
```
jaywink, line 396: Is there a reason for this particular change?

clients/clients.go
```diff
@@ -316,14 +316,35 @@ func (c *Clients) onRoomMemberEvent(client *mautrix.Client, event *mevt.Event) {
...
-			logger.Print("Joined room")
+			return
+		}
+		logger.Print("Joined room")
+		var rID = id.RoomID(event.RoomID.String())
+		var plContent struct {
+			Users        map[string]int `json:"users"`
+			UsersDefault int
+		}
+		if client.config.MinimumPowerLevel == 0 {
```
jaywink, line 333: This should probably be optional?

jaywink (commented):
Some initial comments. Some documentation updates would be needed as well imho.

clients/clients.go
Half-Shot: Go defaults unset integer values to 0, which is incredibly annoying in this context.

clients/clients.go
Half-Shot: So we can check the config, which is on botClient but not client

clients/clients.go
Half-Shot: I did some learning of go, and doing aMap["foo"] will return *two* values. The first is the value, and the second is whether the value existed in the map.

So in this case, I can do `var pl, userExists` and if userExists is false, we fall back to the default PL.

Issue #349, opened 2021-05-13
Github bot: Support automatic labels/projects/etc for issues created in a particular room


turt2live:
May require a github-side bot as not everyone will be able to self-apply labels.

PR #358, opened 2021-08-18
Add support for GitHub labels when creating a new issue

novocaine:
Augmented the current `m.room.bot.options` format for the github integration to support new_issue_labels, a comma separated list of label names that will be added to the issue.

This is useful for triggering subsequent automation from bot-created issues.

This partly addresses https://github.com/matrix-org/go-neb/issues/349 - although that issue notes that not everyone will have permission to add labels. I didn't attempt to address that here; this feature will just be less useful for those particular cases.

### DIFF
--- a/clients/clients.go
+++ b/clients/clients.go
@@ -328,6 +328,13 @@ func (c *Clients) onRoomMemberEvent(client *mautrix.Client, event *mevt.Event) {
 	}
 }
 
+type BotOptionsContent struct {
+	Github struct {
+		DefaultRepo    string `json:"default_repo,omitempty"`
+		NewIssueLabels string `json:"new_issue_labels,omitempty"`
+	}
+}
+
 func (c *Clients) initClient(botClient *BotClient) error {
 	config := botClient.config
 	client, err := mautrix.NewClient(config.HomeserverURL, config.UserID, config.AccessToken)
@@ -344,12 +351,10 @@ func (c *Clients) initClient(botClient *BotClient) error {
 	botClient.verificationSAS = &sync.Map{}
 
 	syncer := client.Syncer.(*mautrix.DefaultSyncer)
-	syncer.ParseErrorHandler = func(evt *mevt.Event, err error) bool {
-		// Events of type m.room.bot.options will be flagged as errors as this isn't an event type
-		// recognised by the Syncer, but we need to process them so the bot can accept options
-		// through this event (see onBotOptionsEvent)
-		return evt.Type.Type == "m.room.bot.options"
-	}
+
+	// Add m.room.bot.options to mautrix's TypeMap so that it parses it as a valid event
+	var StateBotOptions = mevt.Type{Type: "m.room.bot.options", Class: mevt.StateEventType}
+	mevt.TypeMap[StateBotOptions] = reflect.TypeOf(&BotOptionsContent{})
 
 	nebStore := &matrix.NEBStore{
 		InMemoryStore: *mautrix.NewInMemoryStore(),
--- a/clients/clients.go
+++ b/clients/clients.go
@@ -330,9 +330,9 @@ func (c *Clients) onRoomMemberEvent(client *mautrix.Client, event *mevt.Event) {
 
 type BotOptionsContent struct {
 	Github struct {
-		DefaultRepo    string `json:"default_repo,omitempty"`
-		NewIssueLabels string `json:"new_issue_labels,omitempty"`
-	}
+		DefaultRepo    string   `json:"default_repo,omitempty"`
+		NewIssueLabels []string `json:"new_issue_labels,omitempty"`
+	} `json:"github"`
 }
 
 func (c *Clients) initClient(botClient *BotClient) error {
--- a/clients/clients.go
+++ b/clients/clients.go
@@ -16,7 +16,6 @@ import (
 	shellwords "github.com/mattn/go-shellwords"
 	log "github.com/sirupsen/logrus"
 	"maunium.net/go/mautrix"
-	"maunium.net/go/mautrix/event"
 	mevt "maunium.net/go/mautrix/event"
 	"maunium.net/go/mautrix/id"
 )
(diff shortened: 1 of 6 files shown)


services/github/github.go
```diff
@@ -50,17 +51,20 @@ var ownerRepoRegex = regexp.MustCompile(`^([A-z0-9-_.]+)/([A-z0-9-_.]+)$`)
...
 //
 //  {
 //    "github": {
-//      "default_repo": "owner/repo"
+//      // The default repository to use for this room; this allows "owner/repo" to be omitted
+//      // when creating/expanding issues.
+//      "default_repo": "owner/repo",
+//
+//      // Comma-separated Github labels to attach to any issue created by this bot in this room.
+//      "new_issue_labels": "bot-label-1, bot-label-2"
```
turt2live, line 64: it should be possible to define this as an array, then parse it as an array magically in one of the layers. If there's something that's causing that to not work, a comment in `newIssueLabels` would be appreciated for documentation purposes.

turt2live (approved):
Otherwise looks fine to me - sending off to the Integrations Team (which I learned we have) for further review.

clients/clients.go
```diff
@@ -344,6 +344,12 @@ func (c *Clients) initClient(botClient *BotClient) error {
 	botClient.verificationSAS = &sync.Map{}
 
 	syncer := client.Syncer.(*mautrix.DefaultSyncer)
+	syncer.ParseErrorHandler = func(evt *mevt.Event, err error) bool {
+		// Events of type m.room.bot.options will be flagged as errors as this isn't an event type
```
Half-Shot, line 348: This is a bit miserable, but it seems to do the right thing. Presumably there isn't a way in which applications can tell the Syncer how an event type maps to a struct?

services/github/github.go
Half-Shot: +1, especially if you are a dodgy person who puts commas in their issue labels.

services/github/github.go
```diff
@@ -735,24 +742,37 @@ func (s *Service) Register(oldService types.Service, client types.MatrixClient)
...
+			return make(map[string]interface{}), nil
+		} else {
+			err := errors.New("Failed to load bot options")
+			logger.WithError(err).Error(err)
+			return nil, err
 		}
-		return ""
 	}
 	// Expect opts to look like:
 	// { github: { default_repo: $OWNER_REPO } }
```
Half-Shot, line 758: Not anymore. Bit of a shame there isn't a struct for this as well.

services/github/github.go
```diff
@@ -735,24 +742,37 @@ func (s *Service) Register(oldService types.Service, client types.MatrixClient)
...
 	ghOpts, ok := opts.Options["github"].(map[string]interface{})
 	if !ok {
-		logger.WithField("options", opts.Options).Error("Failed to cast bot options as github options")
+		err = fmt.Errorf("Failed to cast bot options as github options")
+		logger.WithField("options", opts.Options).Error(err)
```
Half-Shot, line 762: I don't think you need this?

Half-Shot (changes requested):
Looks excellent, and personally would love to have this. A few nits here and there that I'd like to see if we can solve.

services/github/github.go
```diff
@@ -794,11 +794,15 @@ func (s *Service) newIssueLabels(roomID id.RoomID) []string {
 	if err != nil {
 		return make([]string, 0)
 	}
-	newIssueLabels, ok := ghOpts["new_issue_labels"].(string)
+	newIssueLabels, ok := ghOpts["new_issue_labels"].([]interface{})
 	if !ok {
 		return make([]string, 0)
 	}
-	return strings.Split(newIssueLabels, ",")
+	newIssueLabelsUnboxed := make([]string, 0)
```
novocaine, line 801: I don't actually know go, so if someone knows how to perform this conversion without doing a loop, please shout

clients/clients.go
novocaine: afaict we're stuck with https://github.com/mautrix/go/blob/master/event/content.go#L19, referenced from https://github.com/mautrix/go/blob/master/event/content.go#L154

services/github/github.go
novocaine: The rearrangement here is because I refactored the option loading into `loadBotOptions`, and I'd like to return an error to the caller if it occurs, so created an `err` which I can return .. unless you were suggesting something else

services/github/github.go
turt2live: This is how you do it in Go, sadly. At least I haven't found a better way yet...

services/github/github.go
novocaine: open to suggestions about the struct .. the unmarshalling happens at a higher level of abstraction at https://github.com/matrix-org/go-neb/blob/f71f1a078d469333a4e10a934947eaf0e22e0d2e/database/schema.go#L458 .. so I think it's too late to provide a specifically typed struct by the time we get here..? without a decent refactor

clients/clients.go
tulir: You can just add stuff to `TypeMap`, that's why it's exported

services/github/github.go
tulir: Unmarshaling the options into a struct with the proper types instead of a map might be a better way, but that would also be a more complicated change

clients/clients.go
novocaine: ah okay .. I wouldn't ordinarily dare to mutate a struct owned by a third party library .. !

t3chguy:
Configuration event:

```json
{
  "github": {
    "default_repo": "vector-im/element-web",
    "new_issue_labels": [
      "X-Command"
    ]
  }
}
```

For a user with permissions to set labels for issue creation: https://github.com/vector-im/element-web/issues/21736
And without (no label is added): https://github.com/vector-im/element-web/issues/21739

Half-Shot (approved):
Looks great to me, no nits!

Half-Shot:
That was easy

PR #377, opened 2022-04-13
Bump Go version in Github CI

tadzik:


t3chguy (approved):
:shipit:

tadzik:
I was hoping that the CI will actually check that it fixes things (it does seem to unblock `staticcheck` locally), but apparently its triggers don't include the workflow itself being changed. Is there any way I can force it to run here?

t3chguy:
https://github.com/matrix-org/go-neb/runs/6008727796?check_suite_focus=true cherry-picked your commit over to https://github.com/matrix-org/go-neb/pull/358

tadzik:
Ah, perfect. I guess we can close this one then, with that one going in.

t3chguy:
sgtm, CI passed over in #358 with your change - thanks!

